PR #149, opened 2014-09-16
Use more stdint.h types and document disabling stdint.h

spl:
Summary

* Replaces #145
* Rename `RAPIDJSON_NO_INT64DEFINE` to `RAPIDJSON_NO_STDINT` and revise documentation to address @pah's concern about unconditionally using `int32_t`/`uint32_t`.

As I discussed in #145, I think this change improves several things:

1. Better type documentation
2. The name of the macro `RAPIDJSON_NO_STDINT` more accurately describes its purpose than `RAPIDJSON_NO_INT64DEFINE` did. It disables the `#include <stdint.h>` and doesn't only disable the definition of `int64_t`/`uint64_t`. Of course, this is more significant with the usage of `int32_t`/`uint32_t`.

@pah had the concern that the type definition `int32_t` implied the implementation type would be restricted to 32 bits on a 64-bit system. I'm certain all 64-bit systems will use the most efficient implementation type for `int32_t`. To check, I looked at `stdint.h` on Mac OS X and Linux as well as `msinttypes/stdint.h`: all of them use `int`.

### DIFF
--- a/doc/sax.md
+++ b/doc/sax.md
@@ -62,8 +62,8 @@ using namespace std;
 struct MyHandler {
     bool Null() { cout << "Null()" << endl; return true; }
     bool Bool(bool b) { cout << "Bool(" << boolalpha << b << ")" << endl; return true; }
-    bool Int(int i) { cout << "Int(" << i << ")" << endl; return true; }
-    bool Uint(unsigned u) { cout << "Uint(" << u << ")" << endl; return true; }
+    bool Int(int32_t i) { cout << "Int(" << i << ")" << endl; return true; }
+    bool Uint(uint32_t u) { cout << "Uint(" << u << ")" << endl; return true; }
     bool Int64(int64_t i) { cout << "Int64(" << i << ")" << endl; return true; }
     bool Uint64(uint64_t u) { cout << "Uint64(" << u << ")" << endl; return true; }
     bool Double(double d) { cout << "Double(" << d << ")" << endl; return true; }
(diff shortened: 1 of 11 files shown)


gnzlbg:
Does using `int`, `long`, `short`, ... mean that a file written with rapidjson in one platform might not be readable in another platform?

pah:
@gnzlbg: JSON is a text format, so it can be safely shared across platforms.

gnzlbg:
@Pah but can you accidentally write a 64bit int as ASCII number and then try to read it into a 32bit int in another platform because in one platform int = 64bit and in another int = 32bit?

pah:
After parsing a JSON document, you need to validate the inputs anyway.  And `v.IsInt()` would return `false`, if the numerical value doesn't fit into 32-bit.

gnzlbg:
So does that mean that it is not possible to write 64bit ints?

pah:
Of course you can read/write 64-bit integers, there's `(Is,Get,Set)Int64` as well.
See the [online documentation] (https://miloyip.github.io/rapidjson/classrapidjson_1_1_generic_value.html#a92991a57de8cf42da2e4efc792acefea).

miloyip:
Thank @spl very much for offering this PR. After quite a lot of discussion, I think it is quite a style issue rather than improving in practical situations. As it comes to an official release now, I do not want to potentially break the interface. I think this can be reconsidered for a next major version (which may change the API) in future.

PR #189, opened 2014-10-31
Document traits of types using <type_traits> and static_assert in tests

drewnoakes:
These additional tests state the current traits of `Document`, `Value` and `StringBuffer`. There are slight differences between them. It seems like a good idea to extend this approach across more types, and to review the expected traits across the board.

For now, this piggybacks on `RAPIDJSON_HAS_CXX11_RVALUE_REFS` to determine whether the compiler is C++11.

Note that there are many more traits that could be asserted in tests.

http://en.cppreference.com/w/cpp/header/type_traits

### DIFF
--- a/test/unittest/documenttest.cpp
+++ b/test/unittest/documenttest.cpp
@@ -229,6 +229,31 @@ TEST(Document, UTF16_Document) {
 
 #if RAPIDJSON_HAS_CXX11_RVALUE_REFS
 
+#include <type_traits>
+
+TEST(Document, Traits) {
+  static_assert( std::is_constructible<Document>::value, "");
+  static_assert( std::is_default_constructible<Document>::value, "");
+  static_assert(!std::is_copy_constructible<Document>::value, "");
+  static_assert( std::is_move_constructible<Document>::value, "");
+
+  static_assert(!std::is_nothrow_constructible<Document>::value, "");
+  static_assert(!std::is_nothrow_default_constructible<Document>::value, "");
+  static_assert(!std::is_nothrow_copy_constructible<Document>::value, "");
+  static_assert( std::is_nothrow_move_constructible<Document>::value, "");
+
+  static_assert( std::is_assignable<Document,Document>::value, "");
+  static_assert(!std::is_copy_assignable<Document>::value, "");
+  static_assert( std::is_move_assignable<Document>::value, "");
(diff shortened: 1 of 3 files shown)


thebusytypist:
I have two static assertions failed on Xcode 6 with `-std=c++11 -stdlib=libc++`:

```
static_assert(!std::is_copy_assignable<Value>::value, "");
static_assert(!std::is_nothrow_copy_assignable<Value>::value, "");
```

I suspect that the `Value`(`GenericValue`) is copy assignable.
Check [here](http://en.cppreference.com/w/cpp/concept/CopyAssignable) for the definition of CopyAssignable.

There are two criterions for a CopyAssignable object:
First it should be MoveAssignable. This is true since
```
static_assert(std::is_move_assignable<Value>::value, "");
```
Second it should have a valid `t = v` semantic form.
Let's assume the opposite side: it does not have any valid `t = v` semantic form. 
Then in the function `RemoveMember`(file `document.h`):
```
MemberIterator RemoveMember(MemberIterator m) {
        RAPIDJSON_ASSERT(IsObject());
        RAPIDJSON_ASSERT(data_.o.size > 0);
        RAPIDJSON_ASSERT(data_.o.members != 0);
        RAPIDJSON_ASSERT(m >= MemberBegin() && m < MemberEnd());

        MemberIterator last(data_.o.members + (data_.o.size - 1));
        if (data_.o.size > 1 && m != last) {
            // Move the last one to this place
            *m = *last;
        }
        else {
            // Only one left, just destroy
            m->~Member();
        }
        --data_.o.size;
        return m;
    }
```
The `*m = *last` would fail the compilation.(`m` is a `MemberIterator`, and a `Member` is a `struct` with two `GenericValue`s. I think the `t = v` form is used there.)
This leads to contradiction.
Therefore `Value` does have a valid `t = v` semantic form.

Can you help me verify this,
or please tell me if I am wrong?

pah:
Yes, you can write `t = v` for two (modifiable) `GenericValue`s. But due to the move semantics of RapidJSON, the right-hand-side of the expression is modified during the assignment.

I don't know exactly, how the traits are defined in C++11. Strictly speaking, this behaviour doesn't fulfil the CopyAssignable requirements. But I'm not sure if implementations have to catch this.

PR #192, opened 2014-11-05
CMake support

### LINKED ISSUE #82: Migrate Premake to CMAKE
Due to late official release of premake, many problems cannot be solved easily. For example, VC2012/2013 are not supported yet, Maverick (OS X 10.9) has problem with the linking options, XCode for iOS, etc

I would like to migrate the build system to CMAKE, which is more active and widely used.

jollyroger:
This is my try to add CMake support for RapidJSON and close #82. It was tested and is working for:

 * Debian Linux (w/ libgtest-dev installed and w/ gtest as submodule)
 * Windows XP ( VS2010, Cygwin); Building on MSYS/MinGW give many errors on gtest compilation it seems the last supported version of gtest for MSYS is 1.6.0, although it is not enough for RapidJSON to build using this version since there is no method EqFailure (see https://gist.github.com/jollyroger/45f29f4769cc41c7251f)

Some features available in this pull request:
- [x] In-source and out-of-source builds are supported
- [x] Separate options to build tests, examples and documentation
- [x] Automatic documentation building (if Doxygen is available and respective option is set)
- [x] Smart Google Test Toolkit lookup w/ build options check against Win32/Unix
- [x] Pkgconfig and CMake package lookup and version support (if installed with `make install`)
- [x] Running tests with ctest (using `make test` of `ctest -V` to get verbose output as before)
- [ ] Compilation flags support provided in #126 and #185

As small addition: I fixed linking errors in the documentation.


davidok8:
Hello,

what's the current status of this update? I like this comprehensive PR. I am eager to have this merged into the master branch.
The only missing thing is the compilation flags: would @thebusytypist and @drewnoakes (cf. #126 and #185) be interested in being involved in this PR? I have limited experience with setting compiler flags in CMake but would be happy to help a bit.

For now, I did my own flavor of CMake build system for my urgent needs (cf. https://github.com/davidok8/rapidjson/compare/maint-replace-premake-by-cmake).

drewnoakes:
Does the merging of this PR make #82 and #192 obsolete?

thebusytypist:
Hi @jollyroger,
I have some suggestions on the install process.

Conventionally if a software is going to be installed into a non-default prefix, only `CMAKE_INSTALL_PREFIX` variable have to be altered. And all files(headers, documents, libs, and binaries) are going to be placed under that prefix.

However, in current build of RapidJSON, I need to modify `INCLUDE_INSTALL_DIR`, `LIB_INSTALL_DIR` and `DOC_INSTALL_DIR`. And the variable `CMAKE_INSTALL_PREFIX` is actually not used. (Since all `XXX_INSTALL_DIR`s are `CACHE`ed variables.)

It is OK if this is exactly what you intend for. And it would be better to remind users to change these variables for a custom prefix.

If you instead want to put all those files into the prefix specified by user,
I suggest to not use `CACHE`ed variables.

That is, use
~~~
SET(LIB_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/lib")
~~~
rather than
~~~
SET(LIB_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE STRING "Directory where lib will install")
~~~

In this way all `XXX_INSTALL_DIR` variables won't be displayed for users and hence cannot be edited.

jollyroger:
@thebusytypist, I'm sorry for the late reply. Thank you for your well-thought comment. 

Yes, I intended this behaviour. It covers the common case when install prefix is set up during the first cmake run. This will calculate all other directoris according to their default values that depend on `CMAKE_INSTALL_PREFIX`. Though this won't be done if you want to set `CMAKE_INSTALL_PREFIX` later at some point(with an additional cmake run) and that's what makes your claim valid. However, please consider another case, when you have a few custom `XXX_INSTALL_DIR` variables that don't exactly match the default rules. In this case you would need to set these variables during every cmake run. Your proposed code will regenerate all dependant variables during every cmake run and that's what I wanted to avoid.

PR #213, opened 2014-12-30
Fix two warnings

mlucool:
Change-Id: I555fb66e1576ac96808b498abfb41e07ec319c36

Fixes on clang35 with c++0x:
definition of implicit copy constructor for 'GenericStringRef<char>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated]

definition of implicit copy assignment operator for 'BigInteger' is deprecated because it has a user-declared copy constructor [-Werror,-Wdeprecated]


miloyip:
@mlucool Sorry that I have not checked with this PR for a long time. Do the current version still failing on your platform? If so, can you update this PR or create a new one?
Thank you.

PR #218, opened 2014-12-31
Added a German sentence

patlecat:


patlecat:
Why? :'(

miloyip:
These changes failed some unit tests. You can click on the red cross above to see the details.
Also, a pull request can contain a set of changes. Submitting a file-by-file pull request may pollute the commit history.
By the way, I don't see how this change help to make the project better.

patlecat:
I just thought to add 1 phrase with other special characters as a test would help to make it easier to add a lot more languages after that. I see no point in having those 5 lines only.

The fact that the tests failed with just 1 correct German sentence shows how important it is to add more languages to it!

pah:
@patlecat: Maybe you can investigate the reason for the failing regression tests in your environment?  If it's not just due to breaking the hidden `BOM`s at the beginning of the files, you may have uncovered a bug here (I'm currently inclined to believe that it has been the former, though).

If you can add these sentences to the different files _without_ breaking the regressions, I'm fairly sure that @miloyip will reconsider merging them. But as @miloyip mentioned, a single pull-request would be sufficient, of course.

patlecat:
Sure I can try that out. Don't think that my editor (Notepad++) has added or modified the BOM of the file.

PR #228, opened 2015-01-16
duplicate expression of  is.Take() == 'l'

SeasonLee:


spl:
@SeasonLee It's not a duplicate expression. `is.Take()` is a mutating operation on the `InputStream` `is`. In this case, it's looking for `"null"` in the stream. See the `is.Peek() == 'n'` followed by `is.Take()` and `is.Take() == 'u' && is.Take() == 'l' && is.Take() == 'l'`. That sequence removes `"null"` from the stream.

You can also see how many tests have failed due to this change in the Travis CI run.

SeasonLee:
@spl OK, i see.It's not a duplicate expression.

PR #230, opened 2015-01-24
Minor additon - ASSERT when no memory in AddChunk. 

kaidokert:
Helps with custom allocator debugging


miloyip:
Thank you for your PR.
However, I think the out-of-memory checking can be done within the allocator, instead of the caller.

PR #232, opened 2015-01-24
Use aligned SSE register load intrinsic.

drewnoakes:
The code immediately before this changed line goes to the trouble of ensuring that data is aligned at a 16-byte boundary, then goes ahead and uses the unaligned form of the load intrinsic `_mm_loadu_si128`.

Either the code shouldn't bother aligning the data to the start of the whitespace, or it should use the aligned form of the intrinsic.

It's possible that on real-world JSON, better performance is observed by avoiding the serial walk to the boundary and instead using the unaligned load directly. There's no hard _requirement_ to have data aligned here, as far as I know. Was this tested?

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -277,7 +277,7 @@ inline const char *SkipWhitespace_SIMD(const char* p) {
 
     // The rest of string using SIMD
 	static const char whitespace[16] = " \n\r\t";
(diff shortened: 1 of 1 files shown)


miloyip:
I think your pull request should be correct.

About the second issue, I remembered I have tested and tuned as the current code. If you find other situations, please feel free to tell us.

PR #234, opened 2015-02-02
Add std::string overload to Writer when RAPIDJSON_HAS_STDSTRING defined

drewnoakes:
Currently `RAPIDJSON_HAS_STDSTRING` is only used in `document.h`. It is also useful in `Writer::String` and `PrettyWriter::String`

### DIFF
--- a/example/serialize/serialize.cpp
+++ b/example/serialize/serialize.cpp
@@ -19,7 +19,7 @@ protected:
     void Serialize(Writer& writer) const {
         // This base class just write out name-value pairs, without wrapping within an object.
         writer.String("name");
-        writer.String(name_.c_str(), (SizeType)name_.length()); // Suppling length of string is faster.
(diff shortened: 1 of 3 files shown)


miloyip:
Please also fix the errors shown in Travis CI build. Thank you.

drewnoakes:
@miloyip, build is passing now, thanks.

drewnoakes:
@miloyip, there are other places where this behaviour could be applied (eg. `Member::find`). Would you like me to submit further PR(s) for these? Also (apart from not failing the build!) is there anything else you need in a PR in terms of docs/test/etc?

miloyip:
Firstly, thank you for your contribution.
I think that `Document` should fully support `RAPIDJSON_HAS_STDSTRING` already. If there are some missing cases, you may try to reproduce it in unit tests and drop an issue. The user guide/examples may miss some documentation on this yet.

PR #241, opened 2015-02-14
Compilation fixes

yachoor:
Fixes compilation issues with some less common configurations:
- compiler that does not import c stdlib functions into global namespace when including through c++ headers (cstring vs string.h)
- platform with uint32_t typedefed to 32-bit unsigned long and uint64_t to 64-bit unsigned long long - compiler cannot decide between overloads when using unsigned int literal


pah:
LGTM :+1: 

Which platform are you using? The second bullet is in fact one of my main objections against the adoptions of #149.

yachoor:
I got this error during compilation using arm-elf-gcc 4.3.2 toolchain, targeting ARM922T processor.

Issue #240, opened 2015-02-11
cmake: No (warning) flags passed during build, several unittests missing


pah:
After the merge of #192, some issues popped up in the current regression runs:

 * All objects are build without any specific compiler flags, instead of the previously passed `-march=native -Wall -Wextra -Werror -Weffc++ -Wswitch-default`

* Some unittests are not built
   * bigintegertest.cpp
   * stringbuffertest.cpp
   * strtodtest.cpp

I've not reviewed the state of the Cmake support indetail.  It might be good to rename `travis-doxygen.sh` to `.travis-doxygen.sh` to avoid confusion with this file visible in the main folder.

miloyip:
Sorry that I should review the PR in more detail first. I just found out there are compilation errors in VS 2013 as well. Do u suggest to revert to PR first?

jollyroger:
Hi. It seemed there were no such unittests at the time I wrote that PR. These are easily added. As for build flags — while most are harmless, `-march` flag isn't something I'd like to add because that could be troublesome for package maintainers (to be more precise, Debian maintainers). For example, library built on a i686-capable (packages are built on a modern systems) machine will not run on an i386 target architecture that is supported by Debian due to missing processor instructions.

In case someone wants to build the library himself CFLAGS, CXXFLAGS are supported by cmake out of the box i.e. running `CXXFLAGS=-march native cmake ..` should be fine.

As for reverting my PR: i'm ok either way. If you revert, I'll rebase my PR on top of current master and add warning flags and unit tests. Otherwise I'll just create another PR. I'm not sure if I could help with VS build though but maybe compilation errors will shed some light.

miloyip:
Hi @jollyroger . Thank you for your quick response.

I am not familiar with package issues. Will they include binaries for something like RapidJSON? Since the binaries are only unit tests and examples, I think probably should not be distributed as binary. If this is true, then `-march` should be safe.

I can try to work on the VC warnings/errors tonight.

jollyroger:
@miloyip , here you can see how previous versions of RapidJSON are distributed: https://packages.debian.org/search?keywords=rapidjson&searchon=names&suite=all&section=all

I'm very sorry, I forgot that it's a source-only library, so you are right, `-march` flag should be safe.

I'll look on the changes meanwhile and decide what to do with that PR.

miloyip:
[...]

PR #244, opened 2015-02-19
Fixes for #240

jollyroger:
These changes target all issues mentioned in bug #240.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,10 @@
 !/bin/data
 !/bin/encodings
 !/bin/jsonchecker
+/build
 /doc/html
 /doc/doxygen_*.db
-/thirdparty/lib
-/intermediate
+*.a
 
 # Temporary files created during CMake build
 CMakeCache.txt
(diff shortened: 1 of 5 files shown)


pah:
Somehow the AppVeyor build has still failed even with your changes. I don't see the changes from 2e79af6 reflected in the [build log] [1]. 

@Kosta-Github: Is the AppVeyor environment cached between runs? Do we need to "clean" the build tree before running? 

[1]: https://ci.appveyor.com/project/miloyip/rapidjson/build/0.12.5/job/2c26ped4qvq8x3kx

Kosta-Github:
AppVeyor is doing a clean build each time it runs ( just like Travis). 

I think the camke definition needs to be -D instead of /D ...

Kosta-Github:
-D is also missing on 2 or 3 other cmake config files (e.g., for the examples)...

Kosta-Github:
The VS2012 compiler does not adhere to the standard with respect to `nothrow` support; some of the `static_asserts` in this file https://github.com/miloyip/rapidjson/blame/master/test/unittest/stringbuffertest.cpp#L86 need to be either disabled for `_MSC_VER` (all VS compiler versions) or selectively for VS2012 (preferred solution)...

PR #249, opened 2015-02-23
RapidJSON.pc.in: include dir not set

chhitz:
Variable in wrong format is not replaced

### DIFF
--- a/RapidJSON.pc.in
+++ b/RapidJSON.pc.in
@@ -1,4 +1,7 @@
+includedir=@INCLUDE_INSTALL_DIR@
+
 Name: @PROJECT_NAME@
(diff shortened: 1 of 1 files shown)


pah:
Good catch! I would prefer to explicitly set the `${includedir}` variable in the pkg-config file, though. So the `RapidJSON.pc.in` file could look like (with some additional improvements):
````make
includedir=@INCLUDE_INSTALL_DIR@

Name: @PROJECT_NAME@
Description: A fast JSON parser/generator for C++ with both SAX/DOM style API
Version: @LIB_VERSION_STRING@
URL: https://github.com/miloyip/rapidjson
Cflags: -I${includedir}
````

chhitz:
adapted PR to suggested version

PR #251, opened 2015-02-24
Update reader.h

qRayner:
duplicated Expression in if condition ...


miloyip:
No.
The original code is correct. It checks and consumes "NULL".
You should run the unit tests when try to modify the code.

qRayner:
sry my bad ...

spl:
@qRayner See #228 for an explanation.

@miloyip To avoid this confusion in the future, you may want to add a comment in the code to explicitly point this out.

PR #257, opened 2015-03-10
GenericValue::AddMember<T>: add missing overload (closes #254)

### LINKED ISSUE #254: Embedded JSON objects issue in Linux
Has anybody have trouble creating embedded json files? I am trying to create something like this:
```
{
"test_name": "sampleName", 
"passed": true
"metrics": { "metric1_name": 12, "metric2_name": 10}
}
```
and then write that to a file. It currently works on my mac but when I try it out on a linux machine it screws up the information in the embedded json (the "metrics" portion). More specifically it writes the wrong string in the metric_names. 

Has anybody stumbled into this issue?

Currently the build status says that it is not correctly building in linux. But rapidjson I am using is from like 6 months ago, if that makes any difference.

Thanks!

pah:
As discovered by @felipegb94, there are missing overloads to the `GenericValue::AddMember<T>` template function, taking an explicit `GenericValue&` as a name and accepting arbitrary primitive values.

This patch adds the missing overloads. The `StringRefType` overload is needed to disambiguate the addition of a string literal as value.

Some tests are added to `TEST(Value, Object)` in `valuetest.cpp`.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -954,6 +954,44 @@ public:
         return *this;
     }
 
+    //! Add a constant string value as member (name-value pair) to the object.
+    /*! \param name A string value as name of member.
+        \param value constant string reference as value of member.
+        \param allocator    Allocator for reallocating memory. It must be the same one as used before. Commonly use GenericDocument::GetAllocator().
+        \return The value itself for fluent API.
+        \pre  IsObject()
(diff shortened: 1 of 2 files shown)


pah:
The build on Travis is current broken due to the missing call to `apt-get update`, see  aae07f6, #258.

PR #258, opened 2015-03-10
Skip running perftest on CI debug builds (Appveyer; Travis)

### LINKED ISSUE #240: cmake: No (warning) flags passed during build, several unittests missing

pah:
The runtimes for the CI builds on Appveyor and Travis CI are currently horribly long.  The main reason for this is the inclusion of the `perftest` even under Debug configurations.  This has not been the case before the cmake migration (#192).

This pull-request excludes `perftest` from the CI configurations `appveyor.yml` and `.travis.yml`.

To avoid merge conflicts, I've merged the pending pull-request #244 (which fixes #240) into this pull-request already. Let's see, what the current status of these pending changes is. ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,10 @@
 !/bin/data
 !/bin/encodings
 !/bin/jsonchecker
+/build
 /doc/html
 /doc/doxygen_*.db
-/thirdparty/lib
-/intermediate
+*.a
 
 # Temporary files created during CMake build
 CMakeCache.txt
@@ -20,4 +20,3 @@ Testing
 install_manifest.txt
 Doxyfile
 DartConfiguration.tcl
-/build/*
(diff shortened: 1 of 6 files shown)


pah:
As a summary: This pull-request fails on
 * MSVC 11
 * Clang 32-bit

I haven't looked at the details, yet.  But I think both issues are not related to this pull-request itself, as it just fixes the Cmake setup and the Appveyor/Travis configuration.

PR #279, opened 2015-04-01
Fixes for GCC with -pedantic

pah:
When compiling with `-pedantic`, some additional warnings are reported by GCC.

* Another trailing comma in an enum in `messagereader.cpp`
* The `unsigned __int128` type is an extension, add an attribute to mark it as such.
* Overflow warnings occur in `encodings.h` due to signedness of `char` on some platforms.  
  Hide those for now, as they are all false positives.

No functional changes (intended).


miloyip:
Thanks

PR #293, opened 2015-04-10
Add -Wfloat-equal and fix all derived warnings 

### LINKED ISSUE #287: Warnings with -Wfloat-equal
Anonymous reported that `-Wfloat-equal` in gcc 4.1.2 causes warnings.

miloyip:
Fix #287

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -700,8 +700,10 @@ public:
             return StringEqual(rhs);
 
         case kNumberType:
-            if (IsDouble() || rhs.IsDouble())
-                return GetDouble() == rhs.GetDouble(); // May convert one operand from integer to double.
+            if (IsDouble() || rhs.IsDouble()) {
+                float delta = GetDouble() - rhs.GetDouble();    // May convert one operand from integer to double.
+                return delta >= 0.0 && delta <= 0.0;            // Prevent -Wfloat-equal
+            }
             else
                 return data_.n.u64 == rhs.data_.n.u64;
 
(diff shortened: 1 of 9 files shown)


include/rapidjson/document.h
```diff
@@ -700,8 +700,10 @@ class GenericValue {
             return StringEqual(rhs);
 
         case kNumberType:
-            if (IsDouble() || rhs.IsDouble())
-                return GetDouble() == rhs.GetDouble(); // May convert one operand from integer to double.
+            if (IsDouble() || rhs.IsDouble()) {
+                float delta = GetDouble() - rhs.GetDouble();    // May convert one operand from integer to double.
```
pah: This should probably say `double delta`, right?
Secondly, why not simply
````cpp
    return !(GetDouble() < rhs.GetDouble() || GetDouble() > rhs.GetDouble());
````

File: test/unittest/readertest.cpp [hunk 2]
```diff
@@ -190,7 +191,7 @@ static void TestParseDouble() {
         internal::Double e(x), a(h.actual_); \
         EXPECT_EQ(e.Sign(), a.Sign()); \
         if (fullPrecision) { \
-            EXPECT_EQ(x, h.actual_); \
+            EXPECT_NEAR(x, h.actual_, 0.0); \
```
pah: Here, we are in `fullPrecision`, so we should indeed check for full equality. Also you did disable the warning already.

miloyip: Sadly the warnings still appear with suppression. I think it is because the headers are included firstly than the pragmas. I try to change it with binary representation comparison.

include/rapidjson/document.h
miloyip: Thank you. I mistakenly write `float` instead of `double` as well. But since GetDouble() is not just returning a value and may involve conversion, I do this instead:
~~~cpp
                double a = GetDouble();     // May convert from integer to double.
                double b = rhs.GetDouble(); // Ditto
                return a >= b && a <= b;    // Prevent -Wfloat-equal
~~~
I applied De Morgan's law :)

PR #295, opened 2015-04-11
Fix several -Wfloat-equal related comments in #293

miloyip:


pah:
Looks good, thanks! :-1:

Issue #294, opened 2015-04-10
Writer test segfaulted under valgrind


shindo:
Valgrind version: valgrind-3.10.1
gcc version 4.9.2 20150304 (prerelease) (GCC) 

RapidJSON was built with `cmake -DRAPIDJSON_HAS_STDSTRING=ON -DRAPIDJSON_BUILD_DOC=OFF`.

Valgrind output (error only): http://pastebin.com/ritSZcrq

miloyip:
I think this may be the same issue as https://github.com/miloyip/itoa-benchmark/issues/8
I will merge that fix here.

miloyip:
I have merged the code to master. Can you help verifying it in your platform? Thank you.

shindo:
Yes, this fixes the problem. Thanks.

miloyip:
Thank you for your report and verification.

PR #296, opened 2015-04-11
Try to fix a potential seg fault on some compiler

miloyip:
Merge the fix from https://github.com/miloyip/itoa-benchmark/issues/8


miloyip:
This may relate to #294

Issue #301, opened 2015-04-14
Crash when passing nearly-empty document


RSpliet:
RapidJSON (debian, 0.12~git20141031-1) seems to crash when I pass it as a string the simple document "{ }". Expected is either a graceful failure in `hasParseError()` or a `return false` on subsequent `hasMember()` requests. Instead, the following occurs:
~~~
Program received signal SIGSEGV, Segmentation fault.
1551	        data_.o.members = (Member*)allocator.Malloc(count * sizeof(Member));
(gdb) bt
#0  SetObjectRaw (allocator=..., count=0, members=0x1edc20, this=0x1edc08) at /usr/include/rapidjson/document.h:1551
#1  EndObject (memberCount=0, this=<optimized out>) at /usr/include/rapidjson/document.h:1852
#2  rapidjson::GenericReader<rapidjson::UTF8<char>, rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >::ParseObject<0u, rapidjson::GenericStringStream<rapidjson::UTF8<char> >, rapidjson::GenericDocument<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator>, rapidjson::CrtAllocator> > (this=<optimized out>, is=..., 
    handler=...) at /usr/include/rapidjson/reader.h:467
#3  0x00034662 in Parse<0u, rapidjson::GenericStringStream<rapidjson::UTF8<> >, rapidjson::GenericDocument<rapidjson::UTF8<> > > (handler=..., is=..., this=0xbefff4c8)
    at /usr/include/rapidjson/reader.h:397
#4  ParseStream<0u, rapidjson::UTF8<>, rapidjson::GenericStringStream<rapidjson::UTF8<> > > (is=..., this=0xbefff500) at /usr/include/rapidjson/document.h:1698
#5  Parse<0u, rapidjson::UTF8<> > (str=<optimized out>, this=0xbefff500) at /usr/include/rapidjson/document.h:1774
#6  Parse<0u> (str=<optimized out>, this=0xbefff500) at /usr/include/rapidjson/document.h:1783
#7  Parse (str=<optimized out>, this=0xbefff500) at /usr/include/rapidjson/document.h:1790
#8  Settings::fillByJSON (this=this@entry=0x1edbc0, json="{ }") at /Embedded/src/Util/Settings.cpp:75
[...]

PR #305, opened 2015-04-15
Avoid calling memcpy with NULL pointers

pah:
According to the C/C++ standards, calling `memcpy(NULL, NULL, 0)` is undefined behaviour. Recent GCC versions may rely on this by optimizing NULL pointer checks more aggressively, see [1].

This patch tries to avoid calling std::memcpy with zero elements. As a side effect, explicitly return NULL when requesting an empty block from MemoryPoolAllocator::Malloc.

This may be related to #301. Opening the pull-request as requested by @miloyip.

[1] https://gcc.gnu.org/gcc-4.9/porting_to.html

### DIFF
--- a/include/rapidjson/allocators.h
+++ b/include/rapidjson/allocators.h
@@ -160,6 +160,9 @@ public:
 
     //! Allocates a memory block. (concept Allocator)
     void* Malloc(size_t size) {
+        if (!size)
+            return NULL;
+
         size = RAPIDJSON_ALIGN(size);
         if (chunkHead_ == 0 || chunkHead_->size + size > chunkHead_->capacity)
             AddChunk(chunk_capacity_ > size ? chunk_capacity_ : size);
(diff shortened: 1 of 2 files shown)


miloyip:
I think adding a branch in `SetObjectRaw()` and `SetArrayRaw()` is not a good idea.
They are only called by `EndObject()` and `EndArray()` respectively. And the parameter should never been `NULL`. Adding assertions should be enough.

pah:
Are you concerned about the performance? I have not seen any difference caused by this change.
But in case of empty arrays/objects (`count == 0`, why wouldn't the pointers be NULL, too?

miloyip:
If the object is to prevent calling `memcpy()` when `count == 0`, I think it should be:
~~~cpp
if (count) {
	data_.a.elements = (GenericValue*)allocator.Malloc(count * sizeof(GenericValue));
	std::memcpy(data_.a.elements, values, count * sizeof(GenericValue));
}
else
	data_.a.elements = 0;
data_.a.size = data_.a.capacity = count;
~~~
`values` is always non-null.

pah:
Oh, you're right. `values` comes from the stack. Will adjust the PR tonight.

pah:
PR updated, see ee7522a.

pah:
Another fix, sorry (0c5c153).

miloyip:
I added a test case and also standardize CrtAllocator::Malloc() in 0e8bbe5e3ef375e7f052f556878be0bd79e9062d.

pah:
Thanks! :+1:

PR #311, opened 2015-04-21
Add version macros for RapidJSON

### LINKED ISSUE #310: Add one function to get rapidjson version
Hey guys,

Could we add one function to get current rapidjson version?
Such as const char* rapidjson_version()

Mack

miloyip:
Since RapidJSON is header-only library, I think that macros should be sufficient.

RapidJSON shall become v1.0.0 release after merge.

Fixes #310


miloyip:
This may be the final commit for v1.0.0.
Can @thebusytypist and @pah review of it?

jollyroger:
It is possible to generate version header automatically using CMake.

 * make additional file called "version.h.in"and put all your version-related stuff there, but using CMake variables in several places: <pre><code>#define RAPIDJSON_MAJOR_VERSION @LIB_MAJOR_VERSION@
    #define RAPIDJSON_MINOR_VERSION @LIB_MINOR_VERSION@
    #define RAPIDJSON_PATCH_VERSION @LIB_PATCH_VERSION@</code></pre>

* add something like this to your CMakeLists.txt:<pre><code>configure_file(version.h.in version.h @ONLY)
include_directories(${CMAKE_CURRENT_BINARY_DIR})
install(FILES ${CMAKE_CURRENT_BINARY_DIR}/version.h DESTINATION ${INCLUDE_INSTALL_DIR})</code></pre>

* include `version.h` in `rapidjson.h`

This approach has some cons though: there's no `version.h` file before you run cmake at least once.

miloyip:
Yes. I have thought of this solution but I don't want cmake compulsory for using the library. Can it do another way round? Parse/match the header to cmake? Besides I do not know how those variables in cmake are used in the build process.

spl:
Just wanted to add that it is a very useful feature to be able to use RapidJSON as a header-only library, without needing to compile anything. So, I agree with @miloyip that a build tool should not be required.

jollyroger:
Those variables were just a stub for you to use. Although I tend to have version defined in the build tool, there are possibilities to get version from other sources. One useful option is to use `git-describe` to get version and parse its output. This version will change with every commit and would be a normal `x.x.x` only for a tagged commit though. Sure, we can get version from header files like this:

 * generate C file that writes version into output and exits
 * run it
 * read its outout

I thought about parsing header files directly and it seems it is doable with some more additional effort.

miloyip:
@jollyroger `git-describe` seems not a good idea because the user may get a source package without git information. I don't know if we can use regex to extract the numbers in `rapidjson.h` in CMAKE. But anyway, this is not hurry we can improve and make it more automatically later. Now we can just simply manually set both files.

@pah Thanks for macro simplification advice. However I try to use the same variable names as those in CMAKE for consistency.

PR #320, opened 2015-04-26
Fix warnings when visited via https

guoxiao:
Browser will show a warning when visiting https://miloyip.github.io/rapidjson/ :
> Mixed Content: The page at 'https://miloyip.github.io/rapidjson/' was loaded over HTTPS, but requested an insecure script 'http://rapidjson-doc.disqus.com/embed.js'. This request has been blocked; the content must be served over HTTPS.

This PR fixes this problem, and updates a link to use https. :smile:

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -28,7 +28,7 @@
 
 6. How to install RapidJSON?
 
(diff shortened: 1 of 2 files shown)


miloyip:
Thanks

PR #328, opened 2015-05-05
Fix alignment of 64bit platforms

isanych:
gcc 5 sanitizer was complaining about alignment on 64bit platform:

pbjson/rapidjson/document.h:1248:9: runtime error: member call on misaligned address 0x631000028b6c for type 'struct GenericValue', which requires 8 byte alignment
0x631000028b6c: note: pointer points here
  74 00 be be be be be be  be be be be be be be be  be be be be be be be be  be be be be be be be be

### DIFF
--- a/include/rapidjson/rapidjson.h
+++ b/include/rapidjson/rapidjson.h
@@ -241,7 +241,11 @@
     alignment. User can customize by defining the RAPIDJSON_ALIGN function macro.,
 */
 #ifndef RAPIDJSON_ALIGN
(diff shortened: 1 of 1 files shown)


miloyip:
Note that, this will increase the sizeof(Value) from 20 bytes to 24 bytes on 64-bit platform.
But I think this would be safer in unknown/future 64-bit architecture.
I will also try to optimize it to 16 bytes on some platforms later, where only 48 bits address space is actually used.

PR #334, opened 2015-05-11
Fix CMakeLists for include as a thirdparty in projects

Phyks:
This fix changes the `CMAKE_SOURCE_DIR` variable to relative variables so that rapidjson could be directly included as a submodule in a bigger project, using `add_subdirectory` CMake command.


miloyip:
I would like to invite @jollyroger to review on this. Thank you.

jollyroger:
Looks fine to me. Thank you for your contribution!

PR #333, opened 2015-05-11
fix compile error when RAPIDJSON_HAS_STDSTRING=1

npeer2004:
This fixes the following error when compiling with RAPIDJSON_HAS_STDSTRING=1.

./rapidjson/include/rapidjson/document.h:992:29: error: no matching function for call to ‘rapidjson::GenericValue<rapidjson::UTF8<> >::GenericValue(std::basic_string<char>&)’
./rapidjson/include/rapidjson/document.h:992:29: note: candidates are:
./rapidjson/include/rapidjson/document.h:554:5: note: rapidjson::GenericValue<Encoding, Allocator>::GenericValue(const std::basic_string<typename Encoding::Ch>&, Allocator&) [with Encoding = rapidjson::UTF8<>, Allocator = rapidjson::MemoryPoolAllocator<>, typename Encoding::Ch = char]
./rapidjson/include/rapidjson/document.h:554:5: note:   candidate expects 2 arguments, 1 provided
./rapidjson/include/rapidjson/document.h:548:5: note: rapidjson::GenericValue<Encoding, Allocator>::GenericValue(const Ch*, Allocator&) [with Encoding = rapidjson::UTF8<>, Allocator = rapidjson::MemoryPoolAllocator<>, rapidjson::GenericValue<Encoding, Allocator>::Ch = char]
./rapidjson/include/rapidjson/document.h:548:5: note:   candidate expects 2 arguments, 1 provided
./rapidjson/include/rapidjson/document.h:545:5: note: rapidjson::GenericValue<Encoding, Allocator>::GenericValue(const Ch*, rapidjson::SizeType, Allocator&) [with Encoding = rapidjson::UTF8<>, Allocator = rapidjson::MemoryPoolAllocator<>, rapidjson::GenericValue<Encoding, Allocator>::Ch = char, rapidjson::SizeType = unsigned int]
./rapidjson/include/rapidjson/document.h:545:5: note:   candidate expects 3 arguments, 1 provided
./rapidjson/include/rapidjson/document.h:542:14: note: rapidjson::GenericValue<Encoding, Allocator>::GenericValue(rapidjson::GenericValue<Encoding, Allocator>::StringRefType) [with Encoding = rapidjson::UTF8<>, Allocator = rapidjson::MemoryPoolAllocator<>, rapidjson::GenericValue<Encoding, Allocator>::StringRefType = rapidjson::GenericStringRef<char>, typename Encoding::Ch = char]
[...]

PR #335, opened 2015-05-12
Add Value::XXXMember(...) overloads for std::string

miloyip:
Try to replace the PR #333
Would like to invite @pah to review.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -844,6 +844,12 @@ public:
     template <typename SourceAllocator>
     const GenericValue& operator[](const GenericValue<Encoding, SourceAllocator>& name) const { return const_cast<GenericValue&>(*this)[name]; }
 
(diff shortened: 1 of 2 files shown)


pah:
As mentioned in #333, these functions have been omitted from the original `std::string` support (#107).
It is a good idea to add them now. For completeness, we should add `PushBack` overloads as well, though.

One of the main reasons for omitting `AddMember( ..., const std::string&, ... )` is the explosion of the overload set if we want to support `std::string` as a _key_ parameter as well. But this will likely pop up soon, too. ;-)

LGTM, otherwise. :+1:

miloyip:
Actually I have tried adding some overloads for `AddMember(const std::string&, ...)` but it will causes ambiguity error for some `AddMember(const Ch*, ...)` calls. `PushBack()` also have similar problem.
So I think this is it.

PR #350, opened 2015-05-31
Fix warnings when compiling with clang and -Wimplicit-fallthrough

c0nk:


c0nk:
Please apply this to the v1.0 branch too.

pah:
I don't understand this change and how this fixes anything (in correctly setup environments).

The purpose of the `RAPIDJSON_PARSE_ERROR_EARLY_RETURN` is to allow "skipping" of the return paths in case exceptions are used for error reporting. This change will require the explicit redefinition of `RAPIDJSON_PARSE_ERROR_EARLY_RETURN_VOID` in these cases as well, which is not documented now.

Can you please provide the exact places where the warning is triggered? I would prefer to address these cases explicitly. 

Thanks, Philipp

miloyip:
The places are actually in iterative parser. I revert the merge and fix those warnings.

c0nk:
Are users allowed to redefine RAPIDJSON_PARSE_ERROR_EARLY_RETURN and RAPIDJSON_PARSE_ERROR_EARLY_RETURN_VOID? My change simply replaced:

if (HasParseError()) { return RAPIDJSON_NOTHING; }

with:

return RAPIDJSON_NOTHING;

The parse error is set on the line before so that if is always true. The if expr was tricking clang into thinking fallthroughs were possible.

PR #353, opened 2015-06-04
Fix #349 emscripten alignment issue

### LINKED ISSUE #349: GetDouble not working in emscripten
Hi,

The `GetDouble()` method seems to always return `0` when rapidjson is compiled using Emscripten.

Here is an example code:
``` c++
#include <string>
#include <fstream>
#include <streambuf>
#include <iostream>

#include "rapidjson/document.h"

int main() {
    std::ifstream in_file("config.json", std::ios::in);
    rapidjson::Document json_config_;
    std::string json_string(
            (std::istreambuf_iterator<char>(in_file)),
            std::istreambuf_iterator<char>());
    if (!json_string.empty()) {
        json_config_.Parse(json_string.c_str());
    }

    std::cout << json_config_["HEIGHTMAP_METADATA"]["CELLHEIGHT"].GetDouble() << std::endl;
}
```

Associated `config.json` file is:
```json
{
    "HEIGHTMAP_METADATA": {
        "CELLHEIGHT": 75.0
    }
}
```

When compiling with `gcc` (`g++ test.cpp -I../path/to/rapidjson/include`) it works fine. When compiling with Emscripten (`em++ test.cpp -I../path/to/rapidjson/include --preload-file config.json -o index.html`) it does not work and outputs 0.

Thanks

miloyip:


Phyks:
Unittests are all passing now, see #349 :)

PR #365, opened 2015-06-24
Add missing allocator to uses of AddMember

mloskot:
AddMember requires allocator, this fix keeps all uses of AddMember consistent across the whole tutorial.


miloyip:
thank you.

PR #369, opened 2015-06-26
Add GenericDocument ctor overload to specify JSON type.

mloskot:
It unifies the interfaces with Value where kXXXType can be passed
into constructor.
It enables shortcut that helps to avoid extra SetXXX() call following
construction of a document.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -1748,6 +1748,13 @@ public:
     typedef GenericValue<Encoding, Allocator> ValueType;    //!< Value type of the document.
     typedef Allocator AllocatorType;                        //!< Allocator type from template parameter.
 
+    GenericDocument(Type type, Allocator* allocator = 0, size_t stackCapacity = kDefaultStackCapacity, StackAllocator* stackAllocator = 0) :
+        GenericValue<Encoding, Allocator>(type),  allocator_(allocator), ownAllocator_(0), stack_(stackAllocator, stackCapacity), parseResult_()
+    {
+        if (!allocator_)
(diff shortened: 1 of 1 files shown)


miloyip:
Please add(copy) the doxygen API documentation.

mloskot:
@miloyip Docs updated.

include/rapidjson/document.h
```diff
@@ -1749,7 +1749,22 @@ class GenericDocument : public GenericValue<Encoding, Allocator> {
...
 
     //! Constructor
-    /*! \param allocator        Optional allocator for allocating memory.
+    /*! Creates an empty document of specified type.
+        \param type             Mandatory type of object to create.
+        \param allocator        Optional allocator for allocating memory.
+        \param stackCapacity    Optional initial capacity of stack in bytes.
+        \param stackAllocator   Optional allocator for allocating memory for stack.
+    */
+    GenericDocument(Type type, Allocator* allocator = 0, size_t stackCapacity = kDefaultStackCapacity, StackAllocator* stackAllocator = 0) :
```
pah: I would recommend to mark this constructor as `explicit` to avoid accidentally creating a temporary `GenericDocument` from a `Type` enum value (because all arguments but the first one are optional).

Strictly speaking, the same applies to the "default" constructor below, but is less important as allocator pointers may not be used as frequently in user code.

mloskot:
@pah Good point, I have added the `explicit` specifier.

pah:
:+1: on this one.

miloyip:
thanks

Issue #368, opened 2015-06-26
How to clear-and-minimize a document?


mloskot:
The documentation clearly explains that neither `RemoveAllMembers` (of object) nor `Clear` (an array) deallocate, so they leave capacity unchanged.

What is the equivalent of clear-and-minimize of standard container?
```
std::vector<T>().swap (v); // or clear followed by shrink_to_fit
```

Is there any recommended way different than the following?
```
Document d;
... // build DOM
d.SetObject();  // clear and minimize
```

miloyip:
All `SetXXX()` calls destructor which deallocates data.
`d.Swap(Value(kObjectType).Move())` should work as well.

mloskot:
Thanks!

pah:
It's probably shorter to write
```cpp
Value(kObjectType).Swap(d); // keeps stack and allocator of "d"
```
That said, I think it's useful to add an explicit `Swap()` function to `GenericDocument` as well, allowing to fully free the memory even in case of a `MemoryPoolAllocator`.
```cpp
Document().Swap(d).SetObject(); // minimize and recreate allocator
Document(kObjectType).Swap(d); // after merging #369
```

mloskot:
:+1: for the explicit `Document::Swap`

amilcarlucas:
To clear and minimize the document I have the following code
```
    if (doc->GetAllocator().Size() > 65536) {
        // force a memory garbage collection, if it requires more than one (default 65KB) block
        rapidjson::Document d; // new temp document
        doc->Swap(d).SetObject(); // minimize and recreate allocator
        d.Swap(*doc);
    }
```
I expect that `doc->GetAllocator().Size()` never grows past 65KiB, but it does.
We never destroy the `doc` document, and we constantly modify it's values.

If `Swap()` is not a solution, then what is?

amilcarlucas:
OK, I got the solution now:
```
  rapidjson::Document* d = new rapidjson::Document;
 ...
  while(1) {
    ...
    change the contents of the doc
    ... 
    if (doc->GetAllocator().Size() > ALLOCATOR_MAX_SIZE) {
        // force a memory garbage collection, if it requires more than one (default 65KB) block
        rapidjson::Document d;
        d.CopyFrom(*doc, d.GetAllocator());
        delete(doc);        // delete it to fully dealocate the memory
        doc = new rapidjson::Document;
        doc->Swap(d);       // now put the clean data back in the document
[...]

PR #376, opened 2015-07-03
add GenericDocument<>::Swap

pah:
See #368.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -1810,12 +1810,11 @@ public:
         \see GenericValue::Swap
     */
     GenericDocument& Swap(GenericDocument& rhs) RAPIDJSON_NOEXCEPT {
-        using std::swap;
         ValueType::Swap(rhs);
         stack_.Swap(rhs.stack_);
-        swap(allocator_, rhs.allocator_);
-        swap(ownAllocator_, rhs.ownAllocator_);
-        swap(parseResult_, rhs.parseResult_);
+        internal::Swap(allocator_, rhs.allocator_);
+        internal::Swap(ownAllocator_, rhs.ownAllocator_);
+        internal::Swap(parseResult_, rhs.parseResult_);
         return *this;
     }
 
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -660,6 +660,8 @@ public:
         return *this;
     }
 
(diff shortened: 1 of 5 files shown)


miloyip:
I suggest not to include `algorithm` for swap. As that header is quite big and may increase compilation overhead. How about creating one internally?

Kosta-Github:
Why not using lowercase "swap" for the method name? In that case it could work hand in hand with std::swap() and std::swap(doc1, doc2) would automatically pick up the optimized implementation; otherwise std::swap() would do the unoptimized create-copy-temp-from-b, copy-a-to-b, copy-temp-to-a thing...

pah:
@miloyip Ok, I'll try to have a look.

@Kosta-Github Providing a lower-case `swap` member function would break the consistency of the RapidJSON API function naming conventions.  Instead, we (or even users) could provide a lower-case `swap` in the `rapidjson` namespace, which should achieve the same Thing.

Kosta-Github:
Yeah, I am aware of the inconsistency but still think that easy/efficient interop with the STL is also worth to be considered.

Users most probably will not provide their own `rapidjson::swap()` function, because the usage of `std::swap(doc1, doc2)` already works out-of-the-box, but unfortunately with less-than-optimal performance.

pah:
@miloyip: I have now added a new helper function `internal::Swap()`, which avoids the dependency on the `<algorithm>` header.

@Kosta-Github: On my platform, `std::swap` doesn't work out of the box for `GenericDocument/GenericValue` as both classes are not copyable by default.  If it works on your platform, then probably due to the enabled C++11 move support.  In the latter case, the default `swap` is nearly as efficient than the explicit one.

We could add an "inline friend" function to `GenericValue`/`GenericDocument` to enable the common `swap` idiom:
````cpp
    // add to GenericValue, GenericDocument (analogously)
    friend inline void swap(GenericValue& a, GenericValue& b) RAPIDJSON_NOEXCEPT
       { a.Swap(b); }

   // enables common pattern:
   using std::swap;
   swap(v1, v2); // calls GenericValue's friend through argument-dependent lookup
````
@miloyip, what do you think?

miloyip:
@pah I think both changes are OK for me.

pah:
I've just added the two free-standing friend `swap` functions.

miloyip:
Brilliant~

PR #382, opened 2015-07-08
Add a 'Reserve' method to the output stream concept

richard-fine:
Add a 'Reserve' method to streams that is called by the writer prior to writing.

The idea here is that if you're writing to an in-memory buffer, you'll have to keep growing the buffer as new content is written to it character-by-character. When the length is known ahead of time - as it is for individual nodes - we can communicate that to the stream, and allow it to grow the buffer at most once per node.

The downside of this is that all custom streams will need an empty "Reserve" method, so it's a breaking change, though not a hard one for people to resolve.

### DIFF
--- a/doc/stream.md
+++ b/doc/stream.md
@@ -254,6 +254,9 @@ concept Stream {
     //! \return Number of characters read from start.
     size_t Tell();
 
+    //! Prepare for writing 'size' characters.
+    void Reserve(size_t size);
+
     //! Begin writing operation at the current read pointer.
     //! \return The begin writer pointer.
     Ch* PutBegin();
@@ -272,7 +275,7 @@ concept Stream {
 ~~~~~~~~~~
 
 For input stream, they must implement `Peek()`, `Take()` and `Tell()`.
-For output stream, they must implement `Put()` and `Flush()`. 
+For output stream, they must implement `Reserve()`, `Put()` and `Flush()`. 
 There are two special interface, `PutBegin()` and `PutEnd()`, which are only for *in situ* parsing. Normal streams do not implement them. However, if the interface is not needed for a particular stream, it is still need to a dummy implementation, otherwise will generate compilation error.
 
 ## Example: istream wrapper {#ExampleIStreamWrapper}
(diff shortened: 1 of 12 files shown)


miloyip:
I have this idea before as well.
But in addition to `Reserve()`, can also add a `PutReserved()` (may be need a better name?) kind of thing, so that it does not need to do space checking for outputting each character.

richard-fine:
Yes, that would be a further optimisation. I think it's not required for this change to be useful as it is, though?

richard-fine:
Alternatively: if we could guarantee that Reserve() was ALWAYS called before calling Put(), then we wouldn't need a separate PutReserved() method - it would be possible for implementors to just assume that space is always available (assuming that they reserved it correctly in Reserve()).

miloyip:
I think it will be a good idea to call `Reserve()` even for single character. So that `Put()` in some streams can be optimized.

Besides, will it be a good idea to use free function with template specialization to make  `Reserve()` optional? Similar to `PutN()` defined in `rapidjson.h`:

~~~cpp
//! Put N copies of a character to a stream.
template<typename Stream, typename Ch>
inline void PutN(Stream& stream, Ch c, size_t n) {
    for (size_t i = 0; i < n; i++)
        stream.Put(c);
}
~~~

And its specialization in `stringbuffer.h`:

~~~cpp
//! Implement specialized version of PutN() with memset() for better performance.
template<>
inline void PutN(GenericStringBuffer<UTF8<> >& stream, char c, size_t n) {
    std::memset(stream.stack_.Push<char>(n), c, n * sizeof(c));
}
~~~

So that the implementation of `Reserve()` will become optional, not affecting existing custom streams.

pah:
With some template magic you can also detect statically, if a stream provides a reserve member (with a matching signature). See, e.g. [Boost.MPL](http://www.boost.org/doc/libs/1_58_0/libs/mpl/doc/refmanual/has-xxx-trait-named-def.html).

miloyip:
Superseded in ef6957c177bd6d971bfedfbfe656c4b82d71f9a3, which use namespace scope functions to provide the optimizations, instead of extending the Stream concept. It prevents the change of API of existing custom streams that may exist.
Thank you for your suggestions and efforts.

PR #390, opened 2015-07-20
And description the value object must be in the scope of document.

tanguofu:
And description the value object must be in the scope of document.

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -470,7 +470,7 @@ For removing members, there are several choices:
 `MemberIterator RemoveMember(MemberIterator)` uses a "move-last" trick to achieve constant time complexity. Basically the member at iterator is destructed, and then the last element is moved to that position. So the order of the remaining members are changed.
 
 ## Deep Copy Value {#DeepCopyValue}
-If we really need to copy a DOM tree, we can use two APIs for deep copy: constructor with allocator, and `CopyFrom()`.
+If we really need to copy a DOM tree, we can use two APIs for deep copy: constructor with allocator, and `CopyFrom()`.The value object was allocated memory from the allocator of the docment. So it is only valid in the scope where the document object is available.
 
 ~~~~~~~~~~cpp
 Document d;


doc/tutorial.md
```diff
@@ -470,7 +470,7 @@ For removing members, there are several choices:
 `MemberIterator RemoveMember(MemberIterator)` uses a "move-last" trick to achieve constant time complexity. Basically the member at iterator is destructed, and then the last element is moved to that position. So the order of the remaining members are changed.
 
 ## Deep Copy Value {#DeepCopyValue}
-If we really need to copy a DOM tree, we can use two APIs for deep copy: constructor with allocator, and `CopyFrom()`.
+If we really need to copy a DOM tree, we can use two APIs for deep copy: constructor with allocator, and `CopyFrom()`.The value object was allocated memory from the allocator of the docment. So it is only valid in the scope where the document object is available.
```
mloskot: Typo: s/docment/document/

I think the sentences should read: the value objects memory was allocated by the allocator of the document.

pah:
While I agree that it is helpful to add some documentation about the allocator relationship between `Document` and `Value`, the proposed place is not the best one.

In fact, **copying** (by using the dedicated constructor or the `CopyFrom` member function, as mentioned in the preceding sentence) is the only safe way to transfer elements between two different documents, potentially even across allocator *types*.

It would be better to add a corresponding paragraph to the description of the [move semantics] [1] in RapidJSON (in the same file).

My 2¢,
Philipp

[1]: https://miloyip.github.io/rapidjson/md_doc_tutorial.html#MoveSemantics

mloskot:
Philipp, I speak from RapidJSON user's perspective and I'd expected a dedicated "Memory management" section before I jumped into the one about move semantic or copying. It simply would be natural sequence of topics.

miloyip:
This PR contains incorrect information. An `allocator` parameter must be provided for deep-copy.
For the suggestions, I open another issue for discussion.

Issue #387, opened 2015-07-16
Passing a Document by-value to a function taking a Value is broken in c++11


lukeworthtkbt:
If we have
```
Document TransformValue(Value v);
```

This works (as far as I can tell):
```
Document d;
d.Parse(json);
Document transformed = TransformValue(std::move(d));
```

But this causes really confusing problems:
```
Document ParseDocument(std::string json) {
   Document d;
   d.Parse(json);
   return d;
}

Document transformed = TransformValue(ParseDocument(json));
```

I think it is because the temporary value returned by ParseDocument gets move-constructed into the argument to TransformValue, but the move is incomplete because it retains the allocator, so when that temporary is deleted, weird stuff starts happening to `v` inside `TransformValue`.

pah:
I'd say, this is the expected behaviour. Moving from a `Document` to a `Value` obviously only moves the `Value` part of the `Document`. Where should the allocator etc. be moved to? Consequently, it will stay in the original `Document`.

In your expression, the (temporary) allocator gets destroyed after the full expression; The `transformed` `Document` now contains the allocator of the (internal) `Document`, returned by `TransformValue`. Obviously, this allocator doesn't match the one used to allocate the contents of the incoming `Value`, which had been allocated by the temporary `Document`, returned by `ParseDocument`. So you end up with `transformed` containing references to memory that has been allocated by different allocators, one of which frees its memory right away.

The same happens in your first code snippet. This will likely break as well, depending on the details of `TransformValue`.

Long story short: Never mix values from different (even implicit) allocators across `Document` objects.  Either pass the entire `Document` or at least pass the/an owning allocator as a separate parameter.  You could for instance define two overloads (untested):

````cpp
Document TransformValue(Value v, Document::AllocatorType& alloc) {
  Document d(&alloc);
  v.Swap(d); // "move" v to d
  // original body of the function
  return d;
}
inline Document TransformValue(Document d) {
  return TransformValue(d, d.GetAllocator());
} 
````

hth,
  Philipp

lukeworthtkbt:
Thanks for your response. The problem is that Document is a subclass of Value, which suggests to the user that you can pass a Document where a Value is expected, which leads to using temporary variables like I did in my first example.
Semantically, it seems to me that Document has-a Value and an Allocator, rather than Document is-a Value with an Allocator. I am not the first to suggest this change: https://github.com/miloyip/rapidjson/issues/380#issue-93603826

pah:
[...]

PR #391, opened 2015-07-20
Prohibit C++11 move from Document to Value

pah:
As reported in #387, silently moving a `GenericDocument` to a `GenericValue` can lead to object slicing and premature deletion of the owning allocator of the (surviving) `GenericValue`.

To reduce this risk, prohibit move construction/assignment of a `GenericValue` from a `GenericDocument`.


miloyip:
Thanks

Issue #415, opened 2015-08-31
Add more Parse Api


leopoldfreeman:
I need an api which can parse buffer of specified length, or I have to make a copy. This is required when I received a data frame from network, but only part of the frame is json.
```cpp
void foo(const char * buff, uint32_t len)
{
    Document d;
    d.Parse(buff, len);
    ...
}
```

swisspol:
:+1:

miloyip:
Currently it can be done via:

~~~cpp
MemoryStream ms(buffer, size);
EncodedInputStream<UTF8<>, MemoryStream> eis(ms);
document.ParseStream(eis);
~~~

PR #416, opened 2015-08-31
Add api to parse string with specified length

leopoldfreeman:
[issue #415](https://github.com/miloyip/rapidjson/issues/415)


miloyip:
I suggest creating another stream for this. Because the current implementation reduces performance of existing usage.

leopoldfreeman:
OK.

Issue #158, opened 2014-10-09
Parse non-null-terminated strings / Parse with std::string


spl:
I would like to parse strings that are not null-terminated but do have a string length.

What's the best way to do that now? Implement the `Stream` concept? I guess it would involve something like copying the `GenericStringStream`, adding a length-remaining member, and modifying `Peek()` to return `'\0'` when the end is reached.

I think this would be useful for RapidJSON in general. It could be exposed as overloaded `Parse(const Ch *, size_t)` and `ParseInsitu (Ch *, size_t)` methods.

miloyip:
I am OK for PR for supporting `Parse(std::string)`.
That may be put in v1.1 Beta.

jmrico01:
> This doesn't work with `ParseInsitu` yet or does it?

+1 ! I don't think it does - `MemoryStream` doesn't support the required write ops. Maybe we need an `InsituMemoryStream`?

PR #417, opened 2015-08-31
Add api to parse string with specified length

leopoldfreeman:
Added new BufferStream to parse string.
[issue 415](https://github.com/miloyip/rapidjson/issues/415)

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -1982,7 +1982,7 @@ public:
     template <unsigned parseFlags, typename SourceEncoding>
     GenericDocument& Parse(const Ch* str, size_t len) {
         RAPIDJSON_ASSERT(!(parseFlags & kParseInsituFlag));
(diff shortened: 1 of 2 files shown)


pah:
Great :+1:, this one closes the first third of #158. 

For completeness, an insitu variant (with a `GenericInsituBufferStream`) and an overload for `std::string` could be added.  The latter has been sketched in https://github.com/miloyip/rapidjson/issues/158#issuecomment-59574781 already.

Thanks!

miloyip:
Oh... I really forgot about `MemoryStream`, which should be used instead.
Besides, we need to add the APIs and some unit tests on this.

leopoldfreeman:
Yes, MemoryStream is good enough for my case. I'll add the api in my fork. Thanks. And I'll not send the pull request because it doesn't deal with the encoding. Sorry.

PR #424, opened 2015-09-15
Fix crosscompile: remove -march=native

YannickKiekens:
-march=native does not make any sense when crosscompiling


spl:
Should cross-compiling be a separate build option/configuration? What about when not cross-compiling? See #163 and #164.

YannickKiekens:
performance critical distro's imply -march=avx or -march=corei7 by default on their toolchains
other distros like Debian & ubuntu prefer maximum compatibility and use "older" -march default settings on the toolchain

it is generally frowned upon to hardcode this flag in a build system. 
It breaks crosscompiling and it will result in random segfaults on x86 systems where the host system is more modern then the target system. 
ex:
package build on corei7, installed on older core 2 duo.

spl:
First, RapidJSON is a “header-only” distribution. The configuration is not used for building a RapidJSON library. I believe the build configuration is primarily for

1. running unit and performance tests and
2. demonstrating configuration options users have for building their own software with RapidJSON.

Second, I have shown you where `-march=native` is used and why. If you want to remove `-march=native`, I think you should replace it with a flag or flags that imply `__SSE2__` and `__SSE4_2__` for the platforms that have those options.

miloyip:
I agree with @spl. Please provide the scenario which requires cross-compiling the tests/demos. If so, may add a CMAKE option for that.

PR #431, opened 2015-09-26
Issue 430, implemented std::iostream support.

bruno-j-nicoletti:
I've gone and implemented std::iostream support in the issue-430 branch. This is done in two independent header files. A unit test cpp file has also been added to test them, based on the existing unit test for FILE *.

I've added two new header files and a new unit test. I have modified the unittest.h header file to add a function that names a . It all passes unit tests on my mac quite happily. 

Please have a look and if you are happy pull them onto your master branch.

### DIFF
--- a/include/rapidjson/iostreamreadstream.h
+++ b/include/rapidjson/iostreamreadstream.h
@@ -71,8 +71,8 @@ public:
   //! For encoding detection only, called at the start on openning a stream.
   const Ch* Peek4() const {
     istream_.get(peekBuffer_, 4);
-    int n = istream_.gcount() - 1;
-    for(int i = 0; i <= n; ++i) {
+    std::streamsize n = istream_.gcount() - 1;
+    for(std::streamsize i = 0; i <= n; ++i) {
       istream_.putback(peekBuffer_[n - i]);
     }
     return peekBuffer_;
--- a/include/rapidjson/iostreamreadstream.h
+++ b/include/rapidjson/iostreamreadstream.h
@@ -41,14 +41,17 @@ public:
   {
     // given the Reader Concept, we need to explicitly check for eof and return '\0'
     int v = istream_.peek();
-    if(v == std::char_traits<wchar_t>::eof()) return '\0';
+    if(v == std::char_traits<Ch>::eof()) return '\0';
     return v;
   }
 
   //! Extract the next character in the stream.
   Ch Take()
   {
-    return (istream_.eof()) ? '\0' : istream_.get();
+    // given the Reader Concept, we need to explicitly check for eof and return '\0'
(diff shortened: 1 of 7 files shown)


include/rapidjson/iostreamreadstream.h
```diff
@@ -0,0 +1,88 @@
...
+  void Flush() { RAPIDJSON_ASSERT(false); }
+  Ch* PutBegin() { RAPIDJSON_ASSERT(false); return 0; }
+  size_t PutEnd(Ch*) { RAPIDJSON_ASSERT(false); return 0; }
+
+  //! Get a pointer to the next four characters in the stream.
+  //!
+  //! For encoding detection only, called at the start on openning a stream.
+  const Ch* Peek4() const {
+    istream_.get(peekBuffer_, 4);
+    int n = istream_.gcount() - 1;
```
mloskot: Why `int` instead of `streamsize`?

bruno-j-nicoletti: Because I don't expect it to be bigger than 4. But yes, I'll change it to std::streamsize, along with **i** in the following loop.

mloskot: :+1:

miloyip:
Thank you for your PR.

But first of all, I need to clarify that the copyright owner cannot be changed. Currently the contributors are listed [here](https://github.com/miloyip/rapidjson/graphs/contributors). If all contributors changes the license header for different files, then there will be very trouble to make the MIT license working. You may change the headers in your own fork. Otherwise, I think these additional files can be placed in a `contribute` directory with whatever license.

Secondly, please follow the coding layout of other existing code, including indentation and curly brackets, for consistency.

`istream::get()` and `istream::peek()` should return `istream::int_type`. This may need further explicit conversion. I think using `istream::get(char_type&)` and checking all errors by `if (!istream_.get(c)) return '\0';`, will be better.

bruno-j-nicoletti:
Given that there are multiple contributors to the project, why not create a centralised copyright notification as suggested [here](https://softwarefreedom.org/resources/2012/ManagingCopyrightInformation.html). That way you can acknowledge the copyright of all contributors. That won't get in the way of the MIT licenses in anyway. 

I can tweak the formatting. Do you have explicit guides?

I can flip to the ```C++ istream::get(char_type&)``` call instead.

bruno-j-nicoletti:
Done all that now. With the exception of the copyright headers.

miloyip:
I need further legal consultations from Tencent (copyright holder). That may take some time.

bruno-j-nicoletti:
OK.

miloyip:
Close this since it is supersed by #530
Thank you for your effort.

PR #432, opened 2015-09-26
Writer refactor

bruno-j-nicoletti:
I've gone and done a major refactor on Writer and PrettyWriter. There is now a new base class, **WriterBase**, that both **Writer** and **PrettyWriter** independently inherit from. **Writer** and **PrettyWriter** no longer have any duplicated functions and are much smaller than they were. The only functions left in the old two writer classes are the old **Prefix** function as well as a new function **CloseBlock**. **WriterBase** uses static polymorphism to call the correct implementation of **Prefix** or **CloseBlock**, and no virtuals are involved. The template signature of WriterBase is pretty ugly though, but you now have much less code to do the same job as before

I've also added overloads for any function that took a **'char *'** as an argument and the can now differentiate between a string literal and a non string literal. For example....

``` C++
void writeThing(WriterType &w, const char *value)
{
   myWriter.Key("thing");  // will call the string literal overload and avoid strlen
   myWriter.String(value); // will call the char * overload and not avoid strlen
}
```

With a string literal, the length is known at compile time and so you can avoid calling 'strlen' on them. The code to do that is slightly funky and needs to be templated, but it works. Should make rapidjson even faster for output.

I've also added functor overloads which write key value pairs, make for less verbose client code. So you can now go....

``` C++
void writePerson(WriterType &myWriter, const Person &person)
{
  myWrite.Key(person.uniqueID());      //< key for object
  myWriter.StartObject();              //< start it
  myWriter("age", person.age());       //< person's age, int
  myWriter("name", person.name());     //< person's name, a std::string
  myWriter("height", person.height()); //< person's height, a double
  myWriter.EndObject();                //< close the object
}
```

None of this needs STL or std::strings to work, though I've left the old std::string overloads in there as well as the #ifdef statements to control them. It all passes the current unit tests as well as a few more I've written.

Finally, I've added autoscoping of arrays and blocks. To work best it needs C++11, but can work without (I have a #if  that checks for the C++ version to control it). This is in the new file 'autoblocks.h'. For my last example, you would go...

``` c++
void writePerson(WriterType &myWriter, const Person &person)
{
  rapidjson::ObjectBlock<WriterType> obj(myWriter, person.uniqueID());  /// start an object with the given string as a key
  myWriter("age", person.age());       //< person's age, int
  myWriter("name", person.name());     //< person's name, a std::string
  myWriter("height", person.height()); //< person's height, a double
                                       // object is automatically closed on exit of the scope
}
```

The verbosity of this can be slightly reduced in C++ 11 with use of auto the new **Block** function in **WriterBase**. Useful if you aren't able to have the 'using rapidjson' statement. This is enabled only for C++11 and is of trivial cost and doesn't add any extra dependencies.

``` c++
void writePerson(WriterType &myWriter, const Person &person)
{
  auto obj = myWriter.Block(person.uniqueID());  /// the writer returns the auto block for an object given a key
  myWriter("age", person.age());       //< person's age, int
  myWriter("name", person.name());     //< person's name, a std::string
  myWriter("height", person.height()); //< person's height, a double
                                       // object is automatically closed on exit of the scope
}
```

The automatic blocking for Arrays and Objects has not had a unit test written yet. Tomorrow. The rest is good to go however.

I should also tidy up the comments slightly to bring back some of the doxygen comments I've moved around.

Let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+* Refactored Writer and PrettyWriter to both directly inherit from new class WriterBase and moved all common code into WriterBase. Much less code.
+* WriterBase extends Writer and PrettyWriter by adding overloaded functor methods that write values, ie: bool operator()(TYPE value) where TYPE is one of the types you can write to JSON.
+* WriterBase extends Writer and PrettyWriter by adding overloaded functor methods that write key/value pair. ie bool operator()(char *key, TYPE value) where TYPE is one of the types you can write to JSON.
+* WriterBase has overloaded all functions that take a string as a parameter in such a way that it can differentiate between string literals and char * variables. It avoids a redundant strlen call for string literals as the length is known at compile time.
+* Added new classes ObjectBlock and ArrayBlock. On construction they will start and Object/Array and on destruction close and Object/Array automatically. Reduces bugs and complexity of code by removing the need to remember to close and open Object/Array.
 
 ## [1.0.2] - 2015-05-14
 
@@ -17,7 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 * CMakeLists for include as a thirdparty in projects (#334, #337)
-* Change Document::ParseStream() to use stack allocator for Reader (ffbe38614732af8e0b3abdc8b50071f386a4a685) 
+* Change Document::ParseStream() to use stack allocator for Reader (ffbe38614732af8e0b3abdc8b50071f386a4a685)
 
 ## [1.0.1] - 2015-04-25
 
--- a/include/rapidjson/autoblocks.h
+++ b/include/rapidjson/autoblocks.h
@@ -66,6 +66,10 @@ public :
   }
 protected :
   WRITER *writer_;
+
+private :
+  ObjectBlock(const ObjectBlock&);
+  ObjectBlock& operator=(const ObjectBlock&);
 };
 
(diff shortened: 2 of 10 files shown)


bruno-j-nicoletti:
Gah. Failing in gcc with -Weffc++ because the **WriterBase** class doesn't have a virtual destructor, even though there are no virtuals in any of the code. Passes happily with clang though. Disabling it in the relevant files as was done already and recommitting.

bruno-j-nicoletti:
Should all be done now and ready for review, including unit tests for ObjectBlock and ArrayBlock.

StilesCrisis:
The string literal overloads will output the wrong data for 
    char x[100] = "hello"; w.String(x);

It will put a bunch of \u0000 after the string. 

I attempted a similar optimization and had to roll it back as there was no clean way to differentiate quoted literals from char arrays.

pah:
> I attempted a similar optimization and had to roll it back as there was no clean way to differentiate quoted literals from char arrays.

See [`GenericStringRef`][1] for an implementation that is used in the `GenericValue` API, e g. [`SetString`][2], and others.  The trick used here is to explicitly disallow all conversions from a non-`const` array:
```cpp
 private:
     //! Disallow construction from non-const array
     template<SizeType N>
     GenericStringRef(CharType (&str)[N]) /* = delete */;
```

That said: This is mostly an optimization to avoid the **copy** of the string, not the computation of the its length.  Compilers can optimize away such ` strlen` calls pretty well already today.

[1]: http://rapidjson.org/structrapidjson_1_1_generic_string_ref.html
[2]: http://rapidjson.org/classrapidjson_1_1_generic_value.html#aabb538578a3c1e60c86cda32d062bc1c

PR #439, opened 2015-10-07
Keep Document value unchanged on parse error, fixes #437

### LINKED ISSUE #437: Bug when Parsing an empty object
Hello

First I want to say what a great job you did with this Library,

I found the following bug. The docs states that when parsing if an error occurs then the document object is not changed. How ever I found that when doing the following Parse will nullify the document:
```
    Document d;
    d.SetObject(); // flags are set to 3
    d.Parse("{}"); // Sets flags to 0
```
Thanks

pah:
Keeping the DOM unchanged in case of an error is the intended behaviour according to the [documentation] [1].

Instead of forcing the value to `kNullType` before starting the parsing, store the parsed value upon success via regular move.

[1]: https://miloyip.github.io/rapidjson/md_doc_dom.html#ParseError

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -1886,14 +1886,13 @@ public:
     */
     template <unsigned parseFlags, typename SourceEncoding, typename InputStream>
     GenericDocument& ParseStream(InputStream& is) {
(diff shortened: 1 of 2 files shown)


miloyip:
Thanks

PR #443, opened 2015-10-11
Introduce comments support

andrusha97:
I find support of comments in JSON very useful, especially when one uses JSON as a format for config files. This PR provides a prototype implementation of this feature.
By default comments support is disabled. One should use a new flag `kParseCommentsFlag` to enable the feature, so this implementation is backward compatible and doesn't weaken the standard compliance by default.

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -400,6 +400,7 @@ public:
         ClearStackOnExit scope(*this);
 
         SkipWhitespaceAndComments<parseFlags>(is);
+        RAPIDJSON_PARSE_ERROR_EARLY_RETURN(parseResult_);
 
         if (is.Peek() == '\0') {
             RAPIDJSON_PARSE_ERROR_NORETURN(kParseErrorDocumentEmpty, is.Tell());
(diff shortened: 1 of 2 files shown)


include/rapidjson/reader.h
```diff
@@ -462,6 +463,29 @@ class GenericReader {
...
+    void SkipWhitespaceAndComments(InputStream& is) {
+        SkipWhitespace(is);
+
+        if (parseFlags & kParseCommentsFlag) {
+            while (is.Peek() == '/') {
+                is.Take();
+
+                if (is.Peek() == '*') {
+                    is.Take();
+                    while (is.Take() != '*' || is.Take() != '/') { }
```
miloyip: It seems need to check for null terminator here. Otherwise a malformed input with only open `/*` may read after null terminator.

include/rapidjson/reader.h
```diff
@@ -462,6 +463,29 @@ class GenericReader {
...
+                } else if (is.Peek() == '/') {
+                    is.Take();
+                    while (is.Take() != '\n') { }
```
miloyip: Here as well. If there is no '\n' but '\0', it cannot exit the loop.

andrusha97:
Fixed. Also added some tests to check this bug.

miloyip:
Thank you.

PR #460, opened 2015-10-29
Adding coapp definition

stunney:
To create a nupkg, call CoApp's Write-NuGetPackage .\rapidjson.autopkg
-defines:MYVERSION=1.0.2 (or whatever version you want to call it)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ Testing
 install_manifest.txt
 Doxyfile
 DartConfiguration.tcl
+*.nupkg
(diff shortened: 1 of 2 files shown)


miloyip:
Will it be better using https://github.com/miloyip/rapidjson/blob/master/license.txt as licenseUrl?

stunney:
Yes, yes it would. :)

Sent from Yahoo Mail on Android

Kosta-Github:
Maybe you can add the sample commandline invokation as a comment to the package def file as well?

stunney:
@Kosta-Github  you mean as a comment at the top of the .autopkg file?

stunney:
@Kosta-Github I have added comments to the top of the .autopkg file

PR #478, opened 2015-11-25
fix Document::Parse(const Ch*) for transcoding

haubi:
To allow for an UTF16-Document to Parse(UTF8),
the Parse() argument has to be UTF8-compatible.


miloyip:
Thanks

PR #502, opened 2015-12-29
All parsable number types are now in an anonymous union

thedrow:
This should optimize numbers parsing a tiny bit since we don't need to allocate 3 different variables for the same number to be parsed.
All the existing tests pass on my machine.
I'm still working on the benchmarks though and I'd like some guidance with those.

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -1,5 +1,5 @@
 // Tencent is pleased to support the open source community by making RapidJSON available.
-// 
+//
 // Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip. All rights reserved.
 //
 // Licensed under the MIT License (the "License"); you may not use this file except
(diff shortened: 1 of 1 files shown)


cosinekitty:
I like the idea of optimization, but I'm concerned about this change.

This change looks like it uses a tiny bit less stack space, but I doubt it makes anything measurably faster.  In fact, it probably makes the job of the optimizer harder, possibly resulting in slower code.  Overlapping the memory for mutually exclusive variables like this is the sort of thing C++ compiler optimizers are already very good at, without programmers having to make the code less readable.  So it might actually be slower without using any less stack space for local variables.

This change is also scary in that it doesn't explicitly initialize the `double` and `uint64_t` union members to 0, which are larger (8 bytes) than the `unsigned` field that is initialized (usually 4 bytes).  I didn't spend enough time to prove/disprove correctness to myself, but that reinforces my point:  the fact that it passes unit tests is good, but it makes the code less obviously bug-free and thus harder to maintain and debug later.  Consider someone making a change later without realizing that modifying `d` would mutate the values of `i` and `i64`. Such a change could even pass all the unit tests yet cause bugs in production code.

thedrow:
I checked before making this change.
The union will be zero initialized if the biggest type comes first.
See http://stackoverflow.com/questions/11812555/how-to-zero-initialize-an-union#answer-11812606

thedrow:
Regarding complexity, we can add a comment.
I don't really know what the compiler does. Can we prove this by disassembling the binary?

miloyip:
I think, unless you find out some compilers cannot generate correctly optimized code, we should not do something like this.

1. The compiler shall check the life-cycle of each auto variable, may share their memory.
2. Actually those variables may not allocate stack space at all, as they may just store on registers.
3. This function is at the leave of call-tree, saving stack does not help.

Doing [SSA](https://en.wikipedia.org/wiki/Static_single_assignment_form) by hand is alright for most compilers too. Doing the opposite seems cannot help.

thedrow:
I just checked and the produced ASM is the same.
I guess you learn something new every day,
Thanks :)

PR #503, opened 2015-12-29
Build with 2 threads

thedrow:
This should make the build a bit faster since we have 2 cores on each build VM.
I'm not sure about running the tests in parallel since some of them are performance tests.
What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,8 @@ before_script:
 
 script:
   - cd build
-  - make tests
-  - make examples
+  - make tests -j2
+  - make examples -j2
   - ctest -V `[ "$CONF" = "release" ] || echo "-E perftest"`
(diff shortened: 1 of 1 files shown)


thedrow:
22 minutes instead of 24. Much better.

thedrow:
@miloyip Should the preftests report correct results that can be used for performance evaluation on Travis or are they run just to prove that they work?

miloyip:
I think there is no guarantee on what spec of build machine it will be running, nor number of simultaneous builds that will be running at the same time. So that the results of perftest is not very useful. It merely runs for showing the perftest itself is runnable, without runtime errors.

thedrow:
@miloyip Good to know. The changeset is include in #504 so I'm gonna close this.

Issue #518, opened 2016-01-26
cannot compile `prettyauto.cpp` with windows + mingw


sledgeh:
```
"C:\Program Files (x86)\JetBrains\CLion 1.1\bin\cmake\bin\cmake.exe" --build C:\Users\kyb\.clion11\system\cmake\generated\774cb9e3\774cb9e3\Debug --target all -- -j 4
[ 10%] Built target tutorial
[ 20%] Built target capitalize
[ 30%] Built target messagereader
[ 40%] Built target condense
Scanning dependencies of target prettyauto
[ 50%] Built target pretty
[ 65%] Building CXX object example/CMakeFiles/prettyauto.dir/prettyauto/prettyauto.cpp.obj
[ 65%] Built target simpledom
[ 75%] Built target serialize
Scanning dependencies of target simplereader
Scanning dependencies of target simplewriter
[ 80%] Building CXX object example/CMakeFiles/simplereader.dir/simplereader/simplereader.cpp.obj
D:\devel.emb\workspace\rapidjson\example\prettyauto\prettyauto.cpp: In function 'int main(int, char**)':
D:\devel.emb\workspace\rapidjson\example\prettyauto\prettyauto.cpp:23:27: error: '_fileno' was not declared in this scope
     _setmode(_fileno(stdin), _O_BINARY);    // NEW
                           ^
[ 85%] Building CXX object example/CMakeFiles/simplewriter.dir/simplewriter/simplewriter.cpp.obj
example\CMakeFiles\prettyauto.dir\build.make:62: recipe for target 'example/CMakeFiles/prettyauto.dir/prettyauto/prettyauto.cpp.obj' failed
mingw32-make.exe[2]: *** [example/CMakeFiles/prettyauto.dir/prettyauto/prettyauto.cpp.obj] Error 1
CMakeFiles\Makefile2:1244: recipe for target 'example/CMakeFiles/prettyauto.dir/all' failed
mingw32-make.exe[1]: *** [example/CMakeFiles/prettyauto.dir/all] Error 2
mingw32-make.exe[1]: *** Waiting for unfinished jobs....
[ 90%] Linking CXX executable ..\bin\simplereader.exe
[ 95%] Linking CXX executable ..\bin\simplewriter.exe
[ 95%] Built target simplereader
[ 95%] Built target simplewriter
mingw32-make.exe: *** [all] Error 2
Makefile:137: recipe for target 'all' failed
```

miloyip:
[...]

PR #520, opened 2016-01-27
Fixed  cannot compile `prettyauto.cpp` with windows + mingw #518

sledgeh:
Caption says all. https://github.com/miloyip/rapidjson/issues/518


miloyip:
This will break the gcc build, as shown in Travis CI.

sledgeh:
Seems now (in current *master*) problem does not appear. MinGW compile without errors.

PR #536, opened 2016-02-11
Add Writer::SetMaxDecimalPlaces()

### LINKED ISSUE #362: Truncate Double Values
It'd be nice if the GenericValue::Double() method took a default parameter to truncate double values:

v.Double(12.529999732971191);    ---> 12.529999732971191
v.Double(12.529999732971191, 3) ---> 12.529

miloyip:
Fixes #362

### DIFF
--- a/include/rapidjson/internal/dtoa.h
+++ b/include/rapidjson/internal/dtoa.h
@@ -145,7 +145,7 @@ inline char* WriteExponent(int K, char* buffer) {
     return buffer;
 }
 
-inline char* Prettify(char* buffer, int length, int k) {
+inline char* Prettify(char* buffer, int length, int k, int maxDecimalPlaces) {
     const int kk = length + k;  // 10^(kk-1) <= v < 10^kk
 
     if (length <= kk && kk <= 21) {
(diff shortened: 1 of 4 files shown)


DavidRJohns:
Great new feature.  Great library in general!  Is there a way to add the ability for it to round as opposed to truncating?  I tried to read through the paper with the Grisu2 algorithm and saw that the paper also has a Grisu2b algorithm that rounds.  Is there a way to modify dtoa to use this rounding algorithm?  Maybe both algoritms could be implemented and a class member could be added to distinguish between rounding or truncating.

jdemeyer:
+1 to the comment of @DavidRJohns

Unfortunately, the current implementation is lossy: reading a floating point number with a given number of decimal places and then writing that number back with the same number of decimal places can result in a different number being written. Example: reading `50.907667` and writing it back with 6 decimal places gives `50.907666`.

Llerd:
I needed rounding rather than truncating, so I added this block in the initial part of Prettify function. Use as you see fit.

```
inline char* Prettify(char* buffer, int length, int k, int maxDecimalPlaces) {
  int kk = length + k;  // 10^(kk-1) <= v < 10^kk

  // 29.9995xxx -> 30.000
  auto rounder =  kk + maxDecimalPlaces;
  if (length > rounder && buffer[rounder] >= '5') {
    for (int i = rounder - 1; i >= 0; i--) {
      buffer[i]++;
      if (buffer[i] <= '9')
        break;

      // overflow to next digit
      if (i != 0) {
        buffer[i] = '0';
      } else {
        // 9.9995xxx -> 10.000
        // Need to insert an initial '1', shift decimal and set rounder to 0
        buffer[0] = '1';
        buffer[rounder] = '0';
        kk++;
        k--;
      }
    }
  }
```

Also, tested previous commenters 50.907667 value, but that seems fixed already.

PR #541, opened 2016-02-13
Add stream position check to reader unit tests

jschobben:
Added some checks to the reader unit tests, to make sure the amount of characters read from the input stream is as expected.

I've simply used the current amount of read characters, to make the tests pass.

### DIFF
--- a/test/unittest/readertest.cpp
+++ b/test/unittest/readertest.cpp
@@ -418,7 +418,7 @@ TEST(Reader, ParseNumber_NormalPrecisionError) {
 }
 
 TEST(Reader, ParseNumber_Error) {
-#define TEST_NUMBER_ERROR(errorCode, str, errorOffset) \
+#define TEST_NUMBER_ERROR(errorCode, str, errorOffset, streamPos) \
     { \
         char buffer[1001]; \
         sprintf(buffer, "%s", str); \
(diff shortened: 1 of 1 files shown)


test/unittest/readertest.cpp
```diff
@@ -667,7 +669,18 @@ TEST(Reader, ParseString_Error) {
...
+                streamPos = 3; // 0xC0 - 0xC1
+            else if (c <= 0xDFu)
+                streamPos = 4; // 0xC2 - 0xDF
+            else if (c <= 0xEFu)
+                streamPos = 5; // 0xE0 - 0xEF
+            else if (c <= 0xF4u)
+                streamPos = 6; // 0xF0 - 0xF4
+            else
+                streamPos = 3; // 0xF5 - 0xFF
+            TEST_STRING_ERROR(kParseErrorStringInvalidEncoding, e, 2, streamPos);
```
jschobben: This looks a bit odd, it does not exactly match to section 3.2 of http://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt

jschobben: It matches https://en.wikipedia.org/wiki/UTF-8#Codepage_layout though. If it's on Wikipedia it must be true :)

miloyip:
Thanks

Issue #212, opened 2014-12-27
c++11 integration


reuk:
At the moment, if I try to use a c++11 style for loop (for each key in an object, or for each item in an array) on a rapidjson Value, I get a compiler error telling me that there's no viable 'begin' function:

    /Users/reuben/dev/parallel_raytrace/cmd/main.cpp:244:29: error: invalid range expression of type 'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >'; no viable 'begin' function available
        for (const auto & i : v)

How simple would it be to make this work? Seems to me it'd just be a case of adding `begin` and `end` members which inline down to the existing `Begin` and `End` ones.

Is C++11 compatibility one of the goals of this library? That is, can I ever expect this to work, or should I just use old-style `for (auto i = v.Begin(); i != v.End(); ++i)`?

miloyip:
Done in #542

Issue #162, opened 2014-10-17
Support C++11 range-based loop


drewnoakes:
Currently the `for(auto item& : myValue)` loop is not supported by rapidjson. I guess this is because the case of `Begin` and `End` does not match what the standard expects.

I'd like to use this for array iteration, though I suspect it could also be useful over members as well.

weepy:
+1

evil-jester:
+1

miloyip:
@pah Your idea of `GetArray()`/`GetObject` has been implemented in PR #542 .
Please review the PR when you have time. Thank you.

miloyip:
Merged #542

kklouzal:
As mentioned in the first post, the capitalization of `B` and `E` on `Begin()` and `End()` functions are preventing auto ranged based for looping. for `(auto Val : Doc)`.

I've attempted to change them to lowercase myself however at that point the compiler starts telling you you're trying to access private/protected member functions when you shouldn't be..

pah:
Two things to note here:

* You can use `GetArray()/GetObject()` functions to use range-based `for` over arrays or objects:
* As RapidJSON value are generally not copyable, you should loop over **references** instead

```cpp
  Document doc;
  // ... fill the array
  for( const auto& e : doc.GetArray() ) // loop over array
     std::cout << e.GetInt() << std::endl;

  // ... fill an object
  for( auto& m : doc.GetObject() ) // loop over object (non-const)
    m.value.SetNull(); // modify entry 
```

PR #542, opened 2016-02-14
Templated accessors and range-based for

### LINKED ISSUE #316: Templated value checking and getter
Provide templated functions:

~~~cpp
bool Value::Is<int>() const;
int Value::Get<int>() const;
...
~~~
These functions are useful when the user code has templated types.

A difficulty is that current RapidJSON does not provide adapter classes for array and object. As @pah suggested in https://github.com/miloyip/rapidjson/issues/162#issuecomment-59544950, we can add some class like that. So we can also have `Array Value::GetArray()` and `Array Get<Array>()`, etc.

I create this issue for replacing #153.

miloyip:
Fix #316, #162, #212 

Request for code review

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -2325,8 +2325,8 @@ public:
     typedef GenericArray<false, ValueT> Array;
     typedef ValueT PlainType;
     typedef typename internal::MaybeAddConst<Const,PlainType>::Type ValueType;
-    typedef typename ValueType::ValueIterator ValueIterator;
-    typedef typename ValueType::ConstValueIterator ConstValueIterator;
+    typedef ValueType* ValueIterator;  // This may be const or non-const iterator
+    typedef const ValueType* ConstValueIterator;
     typedef typename ValueType::AllocatorType AllocatorType;
     typedef typename ValueType::StringRefType StringRefType;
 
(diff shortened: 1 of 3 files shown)


include/rapidjson/document.h
```diff
@@ -394,6 +394,127 @@ template <typename T> struct IsGenericValue : IsGenericValueImpl<T>::Type {};
...
+    typedef std::basic_string<typename ValueType::Ch> StringType;
+    static bool Is(const ValueType& v) { return v.IsString(); }
+    static StringType Get(const ValueType& v) { return v.GetString(); }
+    static ValueType& Set(ValueType& v, const StringType& data, typename ValueType::AllocatorType& a) { return v.SetString(data, a); }
+};
+#endif
+
+template<typename ValueType> 
+struct TypeHelper<ValueType, typename ValueType::Array> {
+    typedef typename ValueType::Array ArratType;
```
pah: Typo here: Arra**t**Type?

include/rapidjson/document.h
```diff
@@ -394,6 +394,127 @@ template <typename T> struct IsGenericValue : IsGenericValueImpl<T>::Type {};
...
+    static bool Is(const ValueType& v) { return v.IsArray(); }
+    static ArratType Get(ValueType& v) { return v.GetArray(); }
+    static ValueType& Set(ValueType& v, ArratType data) { return v.SetArray(data); }
+    static ValueType& Set(ValueType& v, ArratType data, typename ValueType::AllocatorType&) { return v.SetArray(data); }
+};
+
+template<typename ValueType> 
+struct TypeHelper<ValueType, typename ValueType::ConstArray> {
+    typedef typename ValueType::ConstArray ArratType;
```
pah: See above

include/rapidjson/document.h
```diff
@@ -394,6 +394,127 @@ template <typename T> struct IsGenericValue : IsGenericValueImpl<T>::Type {};
...
+    static bool Is(const ValueType& v) { return v.IsArray(); }
+    static ArratType Get(const ValueType& v) { return v.GetArray(); }
+};
+
+template<typename ValueType> 
+struct TypeHelper<ValueType, typename ValueType::Object> {
+    typedef typename ValueType::Object ArratType;
```
pah: Do you mean `ObjectType` here?

include/rapidjson/document.h
```diff
@@ -2172,6 +2333,145 @@ GenericValue<Encoding,Allocator>::GenericValue(const GenericValue<Encoding,Sourc
...
+    ~GenericArray() {}
+
+    SizeType Size() const { return ptr_->Size(); }
+    SizeType Capacity() const { return ptr_->Capacity(); }
+    bool Empty() const { return ptr_->Empty(); }
+    void Clear() const { ptr_->Clear(); }
+    ValueType& operator[](SizeType index) const {  return (*ptr_)[index]; }
+    ValueIterator Begin() const { return ptr_->Begin(); }
+    ValueIterator End() const { return ptr_->End(); }
+    const GenericArray& Reserve(SizeType newCapacity, AllocatorType &allocator) const { ptr_->Reserve(newCapacity, allocator); return *this; }
```
pah: As the wrappers are only a plain pointer anyway (i.e. coying is as cheap as it gets), we can return by value here.  This can help compilers with optimization as it reduced one indirection.

But, more importantly: Why is `Reserve` declared `const`?

pah:
:+1: LGTM, thanks!

I haven't seen a conversion from `Generic(Array|Object)` back to `GenericValue`.  Wouldn't this conversion be helpful to move the contents of an Array/Object back into another `Value`?  The only challenge would be to restore the kArrayType|kObjectType invariant after moving from the nested `Value`.  What do you think? Maybe, it's better to add additional (but here non-`explicit`) constructors to `GenericValue`?

include/rapidjson/document.h
miloyip: You are right. It is copy-and-paste error.
The typedef is just for simplifying the function declarations. I think these structs are internal and should not be needed by user directly.

include/rapidjson/document.h
miloyip: Actually, since these member functions does not change `ptr_` but data pointed by `ptr`, they should be all const member function. And as const member function, they can only return const this (for fluent API).
But I just realized I should also provide a non-const member function version for completeness, shouldn't I?

miloyip:
There are `GenericValue::SetArray|Object(Array|Object&)` already.
I do not understand 
> The only challenge would be to restore the kArrayType|kObjectType invariant after moving from the nested Value. 

Maybe a (move) constructor and assignment will be more complete.
How about other similar operations such as `PushBack(Object|Array)`, `AddMember(...)`, etc?

include/rapidjson/document.h
pah: > But I just realized I should also provide a non-const member function version for completeness, shouldn't I?

Non-const overloads are not needed, I think.

> Edit: Humm.. seems const member function returning a non-reference this will be simpler.

Agreed, and as said I would also prefer to return *by-value*.

pah:
> There are `GenericValue::SetArray|Object(Array|Object&)` already.

Yes, but no `GenericValue(Object/Array)` constructors, allowing to directly move an array/object to another value.  With such (implicit) constructors (and assignments), wouldn't it be sufficient for all other uses?  Adding another bunch of overloads to all value-accepting functions sounds like a lot of boiler-plate to me.

````cpp
  GenericValue(Object o) : RAPIDJSON_NOEXCEPT : data_(), flags_(kNullFlag) {
    RAPIDJSON_ASSERT(o.ptr_);
    RawAssign(*o.ptr_);
    o.ptr_->SetObject(); // maintain invariant
  }
````

File: include/rapidjson/document.h [hunk 5]
```diff
@@ -1298,7 +1429,10 @@ class GenericValue {
 
     //! Set this value as an empty array.
     /*! \post IsArray == true */
-    GenericValue& SetArray() {  this->~GenericValue(); new (this) GenericValue(kArrayType); return *this; }
+    GenericValue& SetArray() { this->~GenericValue(); new (this) GenericValue(kArrayType); return *this; }
+
+    //! Set this value with an array.
+    GenericValue& SetArray(Array& a) { return *this = *a.ptr_; }
```
pah: After this call, `a.ptr_` is null, not an (empty) array. We should add an `a.ptr_->SetArray();` here (set for SetObject(Object);` to avoid breaking the invariant.

Secondly, we should check for a.ptr_ not being NULL in an assert, which could happen due to a default constructed `Array|Object` instance.

include/rapidjson/document.h
```diff
@@ -2172,6 +2333,145 @@ GenericValue<Encoding,Allocator>::GenericValue(const GenericValue<Encoding,Sourc
...
+    typedef typename internal::MaybeAddConst<Const,PlainType>::Type ValueType;
+    typedef ValueType* ValueIterator;  // This may be const or non-const iterator
+    typedef const ValueT* ConstValueIterator;
+    typedef typename ValueType::AllocatorType AllocatorType;
+    typedef typename ValueType::StringRefType StringRefType;
+
+    template <typename, typename>
+    friend class GenericValue;
+
+    GenericArray() : ptr_() {}
```
pah: Can we skip the default constructors for `Array|Object`?  We could then change `ptr_` to a reference and guarantee, that we always point to a value. (See comment above).

miloyip:
Thank you @pah

PR #556, opened 2016-02-21
Add parse-by-parts example

### LINKED ISSUE #527: Add ability to parse json by parts
The parser  enhancement  is required to parse json partially.
The following functionality of the lirary would be very helpful: the ability to send to the parser json request divided by separated parts as it is recieved from server/client/etc.
At the moment it has to wait until all the json parts are recieved and only then the complete json can be passed to the parser. It would be great if the parts can be passed to the parser separately.

miloyip:
Fix #527 

I am not familiar with C++11 concurrency, please help review the code. Thank you.


spl:
Thanks for writing this up!

I'm also not familiar with C++11 threads, so I can't help you much there. But I would suggest not using `ass_` as a variable name.

miloyip:
It seems clang on Travis cannot compile due to [libstdc++ issue](https://llvm.org/bugs/show_bug.cgi?id=12893). I need to migrate to new travis first.

Issue #562, opened 2016-02-27
Does not compile with g++ 4.9


oktal:
Hi,

I tried compiling rapidjson with g++4.9 but it fails with some -Weffc++ related errors:

```
[ 46%] Building CXX object example/CMakeFiles/parsebyparts.dir/parsebyparts/parsebyparts.cpp.o
/home/octal/code/rapidjson/example/parsebyparts/parsebyparts.cpp: In instantiation of ‘AsyncDocumentParser<parseFlags>::AsyncDocumentParser(rapidjson::Document&) [with unsigned int parseFlags = 0u; rapidjson::Document = rapidjson::GenericDocument<rapidjson::UTF8<> >]’:
/home/octal/code/rapidjson/example/parsebyparts/parsebyparts.cpp:131:39:   required from here
/home/octal/code/rapidjson/example/parsebyparts/parsebyparts.cpp:20:5: error: ‘AsyncDocumentParser<>::mutex_’ should be initialized in the member initialization list [-Werror=effc++]
     AsyncDocumentParser(Document& d) : stream_(*this), d_(d), parseThread_(&AsyncDocumentParser::Parse, this), completed_() {}
     ^
/home/octal/code/rapidjson/example/parsebyparts/parsebyparts.cpp:20:5: error: ‘AsyncDocumentParser<>::notEmpty_’ should be initialized in the member initialization list [-Werror=effc++]
/home/octal/code/rapidjson/example/parsebyparts/parsebyparts.cpp:20:5: error: ‘AsyncDocumentParser<>::finish_’ should be initialized in the member initialization list [-Werror=effc++]
cc1plus: all warnings being treated as errors
```

```
g++ --version
g++ (Ubuntu 4.9.2-10ubuntu13) 4.9.2
```

```
uname -ra
Linux octal-desk 3.19.0-47-generic #53-Ubuntu SMP Mon Jan 18 14:02:48 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

I'd like to include your library as a "thirdparty" in my framework, but I need it to compile first. The issue above looks pretty easy to fix, I can contribute if you want me to. Otherwise, I'll have to fork or subtree and remove the parts that do not compile I guess :(

Also, here is the hash I'm using:

```
commit 2d5d9363d812886c7ae6ac1238dc8f0c55e4b893
Merge: 4de9ba5 c62a19c
Author: Milo Yip <miloyip@gmail.com>
[...]

PR #563, opened 2016-02-27
#562: example: parsebyparts now compiles with g++-4.9

oktal:
Fixed the constructor of `AsyncDocumentParser` to properly initialize all the fields and added pthread link.


miloyip:
Thanks

Issue #560, opened 2016-02-26
Parsing nulls/bools/ints/doubles as strings


corporateshark:
The SAX parser generates separate events for null/bool/ints/int64/double values and parses them accordingly. This is according to the JSON specification. However, in some applications significant performance boost can be achieved by eliminating the conversion stage and treating all the values as strings. A handler can be boiled down to this form:
```
{
    bool String(const Ch*, SizeType, bool) { return static_cast<Override&>(*this).Default(); }
    bool StartObject() { return static_cast<Override&>(*this).Default(); }
    bool Key(const Ch* str, SizeType len, bool copy) { return static_cast<Override&>(*this).String(str, len, copy); }
    bool EndObject(SizeType) { return static_cast<Override&>(*this).Default(); }
    bool StartArray() { return static_cast<Override&>(*this).Default(); }
    bool EndArray(SizeType) { return static_cast<Override&>(*this).Default(); }
}
```

miloyip:
This is possible but what will be the application of this? Roundtrip will not be possible with this as well.

Null/true/false does not boost performance IMHO, validation is still necessary. However, I have thought of providing a parse option, which skips conversion of numbers (just validates) and generates event to a special handler function, like:

~~~cpp
bool Number(const Ch* SizeType len);
~~~

It is possible to make a roundtrip for that. And user can parse the number only when it is needed.

corporateshark:
We use this when JSON is parsed into an intermediate data format inside the app and this format treats everything as strings eliminating variant data types at tree nodes. 

I have a working implementation with a new parsing flag and a relatively small patch to `ParseNumber()`. Can clean it up and open a pool request.

h46incon:
Writer.RawNumber() will write number as string, is this in design?

corporateshark:
Probably not. I guess the Writer should be symmetrical to the Reader and write `int`s in a sense that it should try converting string to int and then write it.

h46incon:
I am trying to fix it but meet problem similar to #1152 .

h46incon:
Similiar issues about quotes of raw number: #704 , #852

PR #564, opened 2016-02-28
Implemented feature as in #560 (ints/doubles as strings)

corporateshark:
Fast parsing of ints/doubles as strings as described in #560

Note: the null-termination character is never added, always use explicit length

### DIFF
--- a/example/capitalize/capitalize.cpp
+++ b/example/capitalize/capitalize.cpp
@@ -24,7 +24,8 @@ struct CapitalizeFilter {
     bool Int64(int64_t i) { return out_.Int64(i); }
     bool Uint64(uint64_t u) { return out_.Uint64(u); }
     bool Double(double d) { return out_.Double(d); }
-    bool String(const char* str, SizeType length, bool) { 
+    bool RawNumber(const char* str, SizeType length, bool copy) { return out_.RawNumber(str, length, copy); }
+    bool String(const char* str, SizeType length, bool) {
         buffer_.clear();
         for (SizeType i = 0; i < length; i++)
             buffer_.push_back(static_cast<char>(std::toupper(str[i])));
--- a/example/jsonx/jsonx.cpp
+++ b/example/jsonx/jsonx.cpp
@@ -57,6 +57,13 @@ public:
         return WriteNumberElement(buffer, sprintf(buffer, "%.17g", d));
     }
 
+    bool RawNumber(const char* str, SizeType length, bool) {
+        return
+            WriteStartElement("number") &&
+            WriteEscapedText(str, length) &&
+            WriteEndElement("number");
+    }
+
     bool String(const char* str, SizeType length, bool) {
         return
             WriteStartElement("string") &&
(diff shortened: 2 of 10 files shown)


miloyip:
I think:
1. `BaseReaderHandler::Number()` should call `String()` instead of `Default()` (Similar to `Key()`).
2. Why `kParseNumbersAsStringsFlag` can only be enabled with `kParseInsituFlag`?
3. Is it also possible to eliminate some code path in `ParseNumber()` when this flag is on (to improve performance)? The bottomline is that it can still validate the JSON standard of number.

corporateshark:
Some code fragments can be eliminated without breaking the validation rules but it does not seem much can be done. For example, code like this can be excluded:

```
            if (expMinus)
                exp = -exp;
```

However, it is not in any tight loop and disabling anything from the parsing code will break JSON validation rules (may be a subject to yet another flag but not this one).

It seems, the encoding unit tests are broken because of this line:
```
typename InputStream::Ch *head = is.PutBegin();
```
What is the proper way to treat encoding of input & output streams?

miloyip:
As example, the following code involving `i`, `significandDigit`, etc. should not be necessary for `kParseNumbersAsStringsFlag`:

~~~cpp
        // Parse int: zero / ( digit1-9 *DIGIT )
        unsigned i = 0;
        uint64_t i64 = 0;
        bool use64bit = false;
        int significandDigit = 0;
        if (RAPIDJSON_UNLIKELY(s.Peek() == '0')) {
            i = 0;
            s.TakePush();
        }
        else if (RAPIDJSON_LIKELY(s.Peek() >= '1' && s.Peek() <= '9')) {
            i = static_cast<unsigned>(s.TakePush() - '0');

            if (minus)
                while (RAPIDJSON_LIKELY(s.Peek() >= '0' && s.Peek() <= '9')) {
                    if (RAPIDJSON_UNLIKELY(i >= 214748364)) { // 2^31 = 2147483648
                        if (RAPIDJSON_LIKELY(i != 214748364 || s.Peek() > '8')) {
                            i64 = i;
                            use64bit = true;
                            break;
                        }
                    }
                    i = i * 10 + static_cast<unsigned>(s.TakePush() - '0');
                    significandDigit++;
                }
~~~

For the following line:
~~~
typename InputStream::Ch *head = is.PutBegin();
~~~

It should only be possible for insitu stream. For other streams, you cannot get pointers to the current position (such as a file stream). 

I think you should make use of existing `NumberStream` to backup the characters in a stack, with modification for `.`, `e` and exponent digits. 

And I will suggest appending `\0` as in `String()` event.

miloyip:
Besides, I think new unit tests are necessary as well.

corporateshark:
Do you mean something like this?
```
        if (parseFlags & kParseNumbersAsStringsFlag) {

            if (parseFlags & kParseInsituFlag) {
                ...
            }
            else {
                const char* str = s.Pop();
                SizeType length = static_cast<SizeType>(s.Length()) - 1;
                cont = handler.RawNumber(str, SizeType(length), true);
            }

        }
```

corporateshark:
Unit tests added.

corporateshark:
Zero-termination in in-situ mode seems problematic. Consider this valid json fragment:

```
					"BufferWidth": 512,
					"BufferHeight": 512,
```

Inserting `\0` after `512` will erase the comma making any further parsing invalid. Currently no zero-character is inserted in in-situ mode, this is explicitly stated in the documentation. Does it sound like a viable solution?

miloyip:
Yes, you are right. there is no `"` for numbers which can be replaced by null-terminator in in situ mode.

miloyip:
Thank you for your contribution.

PR #576, opened 2016-03-06
[simplewriter.cpp] show generated output

ChrisLundquist:
This makes it painfully obvious that writer expects subsequent key/value pairs.


miloyip:
Actually Writer added `Key()` function which can make the example clearer.

ChrisLundquist:
Well, it seems like merging this is pretty harmless until the example is
updated.
My main goal was showing the sample output :)

PR #577, opened 2016-03-08
PrettyWriter formatting options.

comand:
This change adds PrettyWriter::SetFormatOptions with a corresponding
bitfield enum PrettyFormatOptions. This allows options affecting the
format of the PrettyWriter to be set. The first option to be provided
is kFormatSingleLineArray, which instructs the PrettyWriter to write
arrays on a single line, rather than breaking them up onto a line
per element.

### DIFF
--- a/include/rapidjson/prettywriter.h
+++ b/include/rapidjson/prettywriter.h
@@ -24,6 +24,14 @@ RAPIDJSON_DIAG_OFF(effc++)
 
 RAPIDJSON_NAMESPACE_BEGIN
 
+//! Combination of PrettyWriter format flags.
+/*! \see PrettyWriter::SetFormatOptions
+ */
+enum PrettyFormatOptions {
+    kFormatDefault = 0,         //!< Default pretty formatting.
+    kFormatSingleLineArray = 1  //!< Format arrays on a single line.
+};
+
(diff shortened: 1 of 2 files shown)


miloyip:
Thank you.

Xilinxx:
I do love the SetFormatterOptions for a 1D array!

But it needs a little improvement for multidimensional arrays.

Now I come up with a line like this  
`
{ "laserruntimes": [[[32411, 2, 0, 0, 0, 0, 0, 0, 0, 0, 27, 311, 0, 0, 0, 0],[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 24, 0, 0, 0, 0...
}
`
instead of a 1 column output without the kFormatSingleLineArray. (Default)

I was looking at the code but can't figure out how I can know if I am in a nested array, my idea is to change     **EndArray(SizeType memberCount = 0)**  implementation.
`

        if (!empty && !(formatOptions_ & kFormatSingleLineArray)) {
            Base::os_->Put('\n');
            WriteIndent();
        }
        if (!empty && (formatOptions_ & kFormatSingleLine2DArray)) {            
            //How to detect if Array in array?
            Base::os_->Put('\n');            
        }        


    `

PR #582, opened 2016-03-13
do potentially precision-losing conversions explicitly

ktrushin:
Some type conversions which may lead to precision loss especially on 64-bit platforms are done explicitly. As a result, including rapidjson as a thirdparty into another project which uses -Wconversion flag (for g++ and clang) doesn't break compilation of that project.


miloyip:
Thank you

Issue #36, opened 2014-06-30
Strict/Relaxed JSON syntax


miloyip:
Currently RapidJSON is tried to be as strict as possible to the JSON spec.
There are several requests in adding relaxed JSON syntax for parsing. Here lists some possible ones.

Root JSON type
- [x] A document with types other than object and array. (Done in #101)

Comment
- [x] Single line JS Comment (#443)
- [x] Multiple line JS Comment (#443)
- [ ] Comment handler event
- [ ] Comment stored in Value

Object
- [ ] Key without quotes (limited to a sequence of confined character sets, e.g. `(A-Za-z\-_)+[0-9A-Za-z\-_]*`
- [x] extra comma at the end (#584)

Array
- [x] extra comma at the end (#584)

String
- [ ] Single quote pair

Number
- [ ] Fractions begins with `.`, e.g. `.123`
- [ ] Number with '.' but without fraction digits, e.g. `0.`, `123.`
- [ ] Hexadecimal integer, e.g. `0xDEADBEEF`
- [X] `Infinity` and `NaN` (#641)

These relaxed syntax may be useful for some applications. However, this may increase complexity of code base.

Would like to gather opinions on this.

andrusha97:
Hi. Is this issue still active? I'd like to have support of comments because I use RapidJson to parse config files.
May I help with the implementation to speed-up the process?

M1chae1:
If you just want to trim out the comments in JSON file before load into rapidjson, you can reference jsmin class used in the below open source project.

https://code.google.com/p/page-speed/source/browse/lib/trunk/src/third_party/jsmin/cpp/?r=339

Then using MemoryStream to load the minified string and parse the stream with rapidjson::Document.ParseStream to do so.

andrusha97:
@M1chae1, I wouldn't like to use this method because:
1) Double parsing is a kludge, and continuing using my fork of rapidjson with comments support is better than using a kludge.
2) The most important thing: currently, if a config is invalid, my parser reports line number and column of the error. With this minifier it will be impossible.

miloyip:
@andrusha97 This issue is still active. I am now preparing for 1.0 release and these features will be added afterwards. If you would like to, you may implement your solution and I will adapt or take it as a reference.

ErikFr:
[...]

PR #584, opened 2016-03-20
Added optional support for trailing commas

ludocode:
This adds `kParseTrailingCommasFlag` to the reader to allow a trailing comma at the end of maps and arrays. This is part of issue #36, adding optional support for relaxed JSON syntax.


miloyip:
Thank you

PR #589, opened 2016-03-28
RawNumber Improvements / Clang Test Fix / Whitespace Cleanup

jnicholls:
RawNumber parsing with non-insitu streams had a lot of issues.  For starters, only copy optimized streams worked, whereas non-copy optimized streams (such as BasicIStreamWrapper) would not work at all.  This is fixed by using a StringStream for the source stream during the encoding phase just before RawNumber is invoked.

For all non-insitu streams, many types of numbers were not parsed into the temporary StackStream properly, particularly negative and scientific numbers.  I created a new ConsumePush method that does the same thing as Consume, except it calls TakePush() instead of Take(), to grab such chars as '.', 'e', '-', etc.

My editor cleans up trailing whitespace as well, so a lot of diffs are from removing trailing whitespace, which I think is a good thing.  I broke those changes up into their own commits, but I believe they're worthwhile including into this PR.

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -1,5 +1,5 @@
 // Tencent is pleased to support the open source community by making RapidJSON available.
-// 
+//
 // Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip. All rights reserved.
 //
 // Licensed under the MIT License (the "License"); you may not use this file except
@@ -7,9 +7,9 @@
 //
 // http://opensource.org/licenses/MIT
 //
-// Unless required by applicable law or agreed to in writing, software distributed 
-// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR 
-// CONDITIONS OF ANY KIND, either express or implied. See the License for the 
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
 #ifndef RAPIDJSON_READER_H_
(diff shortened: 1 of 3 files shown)


jnicholls:
CI failed, but it's not apparent at all why.  All tests run and pass on my end with both gcc and clang.  Some of the builds on travis didn't even run the tests, claiming the binary couldn't be found 0_o

jnicholls:
CI is hung on VS builds, but I'm sure they'll be okay.

One thing that may be more performant is to avoid the additional branching on `fullPrecision` and instead enhance the NumberStream template class to treat `Take()` to be analogous to `TakePush()` in a specialization, and remove the split decisions on Consume/ConsumePush and Take/TakePush everywhere.  I'll draft that up as an additional commit.

jnicholls:
This latest commit is actually the best choice, and it actually improves performance for number parsing slightly by removing the runtime branching when deciding to Push a period ('.') or a scientific exponent ('e') char.  This seems to be the cleanest solution and is my final work I'd say.

jnicholls:
I rebased everything into a single commit, to keep the history clean and eliminate evidence of backtracking in the design of the fix.

miloyip:
Hi, thank you for your contribution.
Regarding to the double-promotion warning, shall we disable it by changing the code, or adding pragma, instead of command-line suppression? Since it will still be a problem when people including the headers from their project.

jnicholls:
Well the errors only occur when building the Value tests because in those tests you have explicit float literals being passed in to GenericValue.  If a user doesn't explicitly use floats, or they explicitly cast floats into doubles, then they'll never see these warnings.

If you wanted to avoid warnings even in those cases, you'd have to add some float template function specializations to GenericValue and static_cast those floats into doubles.  IMHO, you should let the end user dictate what should be warned or ignored based on their usage.  I only ignored those warnings in your unittests project.

miloyip:
I see. If it is only in unit test code it is fine. I will try to check later to see whether the tests can avoid them.

PR #600, opened 2016-04-08
fix broken links to docs

bogolt:


miloyip:
This is the way to make it works in doxygen-compiled [HTML](http://rapidjson.org/md_doc_tutorial.html).
Thank you anyway.

PR #607, opened 2016-04-18
example checks for integer overflow via RAPIDJSON_ALIGN rounding

struct:
MemoryPoolAllocator::Malloc and MemoryPoolAllocator::Realloc contain potential integer overflows via the rounding macro RAPIDJSON_ALIGN. This example patch checks for that. Its been tested with RapidJson unit tests and all passed. This is just one example of how to do it, YMMV.

### DIFF
--- a/include/rapidjson/allocators.h
+++ b/include/rapidjson/allocators.h
@@ -177,7 +177,14 @@ public:
         if (!size)
             return NULL;
 
+        size_t save_sz = size;
+
         size = RAPIDJSON_ALIGN(size);
+
+        if(save_sz > size || size == 0) {
+            return NULL;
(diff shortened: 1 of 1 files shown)


miloyip:
Hi, may I know does this really necessary? It seems on 32-bit architecture this will only happens if newSize is >= 2^32 - alignment. Should it be guarded by assertion only?

struct:
Its think its necessary if the size value passed to Malloc is ever user controlled and not influenced by the actual size of an input (e.g. a JSON string). This would result in too small of an allocation and an eventual out of bounds write. That isn't the case in the code today. I think an assert works fine.

PR #616, opened 2016-04-25
Define RAPIDJSON_HAS_CXX11_RVALUE_REFS directly in clang

brucestephens:
This makes no difference except that it avoids "warning: macro expansion
producing 'defined' has undefined behavior" messages.


brucestephens:
I'm an idiot...

PR #631, opened 2016-05-09
npm compliant package.json

milkandsour:
In order to be able to install rapidjson via npm, I added a package.json and an include_dirs.js as a main.

I use rapidjson  for my own project and I find very useful to install it from npm, also you will have more visibility if you actually publish the package on the npm repo.

### DIFF
--- a/include_dirs.js
+++ b/include_dirs.js
@@ -1,0 +1,2 @@
+console.log(require('path').relative('.', __dirname));
+
--- a/include_dirs.js
+++ b/include_dirs.js
@@ -1,2 +1,2 @@
-console.log(require('path').relative('.', __dirname));
-
+var path = require('path');
+console.log(path.join(path.relative('.', __dirname), 'include'));
--- a/package.json
+++ b/package.json
@@ -13,13 +13,13 @@
   },
   "repository": {
     "type": "git",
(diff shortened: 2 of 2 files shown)


miloyip:
I have no idea about NPM for C++ library. Is it suitable?

milkandsour:
Yes it is!
In order to create your own nodejs native extension (addon), you need to
write c++ code and that is because nodejs is written on top of v8, that is
a c++ engine.
Native exension are not so common, but if you seek performance, that is the
way, legacy c++ code also could be the case to go native.
Npm give you the ability to use the install command to compile and install
addons (via node-gyp), or just include it, as I did for rapidjson, via
package.json.
Rapidjson it is the fastest, cause the sax skills, I did not run a
comparison but I think it will kill the v8 JSON native.

Anyway, I do not want to force you, tha was just a pull request shaped
suggestion, I did my own fork and,
in case I will decide to use the library:
I will certainly make the necessary endorsements
I will use it "for Good, not Evil" (cit. license)

2016-05-10 6:22 GMT+01:00 Milo Yip <notifications@github.com>:

miloyip:
OK. Thank you.

gggin:
@milkandsour  
how to use this packjson in npm
will you give us a document?

Issue #644, opened 2016-05-30
Undefined behaviour


efidler:
rapidjson relies on undefined behaviour in C++ in several places. I've been trying to make my code that relies on rapidjson run cleanly through ubsan (http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) on Clang/OSX and GCC5/Linux.

I'll put up some fixes as pull requests, but I've got some questions about your preferences.

- Do you want each small fix split out?
- Are you ok with using std::numeric_limits?
- Are you ok with compiler annotations to mark deliberate uses of undefined behaviour?
- Are you ok with using compiler intrinsics for checked arithmetic (at least on Clang and GCC)?

miloyip:
>- Do you want each small fix split out?
I am not sure. But separated fixes should be safer and easier for discussion.

>- Are you ok with using std::numeric_limits?
I try to reduce number of header files include. As `cstdint` cannot be prevented, I try to use those macro limits. But if there is template, maybe really need `limits`.

>- Are you ok with compiler annotations to mark deliberate uses of undefined behaviour?
Ideally try to use macro instead of direct use of annotation.

>- Are you ok with using compiler intrinsics for checked arithmetic (at least on Clang and GCC)?
Not sure about this. Maybe similar to the above one.

efidler:
Ok, the pull request is #646.

I split out the individual fixes. I guess one pull request is ok?

I had to use limits, since cstdint wasn't giving me UINT64_MAX or INT64_* on non-C++11 builds.

I put an #if 0 around a couple of tests in valuetest.cpp because they're undefined by nature and I'm not sure what you're trying to test. I fixed the other "deliberate" undefined behaviour.

I ended up removing the __builtin_mul_overflow use, so I didn't need the intrinsic.

miloyip:
Thank you again for your contributions on this.

PR #646, opened 2016-05-31
Fix undefined behaviour

efidler:
Fix issues with undefined behaviour, mostly found by running the unittests under UBSAN on Clang/OSX and GCC5/Linux.

Issue #644


include/rapidjson/document.h
```diff
@@ -952,12 +956,16 @@ class GenericValue {
         if (IsUint64()) {
             uint64_t u = GetUint64();
             volatile double d = static_cast<double>(u);
-            return static_cast<uint64_t>(d) == u;
+            return (d >= 0.0)
+                && (d < (static_cast<double>(std::numeric_limits<uint64_t>::max()/2 + 1)) * 2)
```
miloyip: Can you explain about this?

include/rapidjson/document.h
```diff
@@ -52,6 +53,9 @@ RAPIDJSON_DIAG_OFF(effc++)
 
 RAPIDJSON_NAMESPACE_BEGIN
 
+static const double LOWEST_FLOAT_AS_DOUBLE = static_cast<double>(-3.4028234e38f);
```
miloyip: Global static variable needed to be prevented, because it may be included by multiple implementation files. In other headers I use inline function to return the values instead, if they needed to be named.

include/rapidjson/internal/dtoa.h
```diff
@@ -102,7 +102,8 @@ inline void DigitGen(const DiyFp& W, const DiyFp& Mp, uint64_t delta, char* buff
         kappa--;
         if (p2 < delta) {
             *K += kappa;
-            GrisuRound(buffer, *len, delta, p2, one.f, wp_w.f * kPow10[-static_cast<int>(kappa)]);
+            int index = -static_cast<int>(kappa);
+            GrisuRound(buffer, *len, delta, p2, one.f, wp_w.f * (index < 9 ? kPow10[-static_cast<int>(kappa)] : 0));
```
miloyip: Was this cause array out of bound?
Otherwise should change to `size_t` and `assert(kappa <= 0)`?

include/rapidjson/schema.h
```diff
@@ -413,9 +413,11 @@ class Schema {
                 }
             }
 
-        AssignIfExist(allOf_, *schemaDocument, p, value, GetAllOfString(), document);
-        AssignIfExist(anyOf_, *schemaDocument, p, value, GetAnyOfString(), document);
-        AssignIfExist(oneOf_, *schemaDocument, p, value, GetOneOfString(), document);
+        if (schemaDocument) {
```
miloyip: This seems unncessary?

include/rapidjson/schema.h
```diff
@@ -578,7 +580,9 @@ class Schema {
     }
 
     ~Schema() {
-        allocator_->Free(enum_);
+        if (allocator_) {
```
miloyip: `allocator_` must not null.

include/rapidjson/document.h
efidler: Casting a double that's out-of-range for the uint64_t is undefined behaviour, so this equality test has undefined results. I'm not 100% clear on what exactly IsLosslessDouble promises if it returns true, but it's certainly possible that you could get true from the equality test when you expect false.

include/rapidjson/document.h
efidler: I guess it doesn't need to be named. I'll just put in the constant.

include/rapidjson/internal/dtoa.h
efidler: I added a test case that causes index to be 13. It's an overflow, not an underflow.

include/rapidjson/schema.h
efidler: allocator_ is NULL for the Schema returned by Schema::GetTypeless(). The destructor is dereferencing the NULL when the global static is torn down during shutdown.

include/rapidjson/schema.h
efidler: schemaDocument is NULL for the Schema returned by Schema::GetTypeless().

include/rapidjson/document.h
miloyip: I am not sure about the calculation `(max / 2 + 1) * 2`. I think it can simply construct a maximum double that can be converted to `uint64_t`.

efidler: The issue is UINT64_MAX can't be exactly represented in a double. I changed it to be strictly less than UINT64_MAX as a double and added some more tests.

efidler:
any more concerns? is this ready to merge?

efidler:
it looks like travis is failing to create the clang toolchain in the VM. I don't think it has anything to do with the code itself.

efidler:
Yes, LLVM removed their apt mirror, which broken travis builds for tons of projects. We could move to the Trusty Travis environment and use the built-in clang, which is 3.5.

miloyip:
Thank you. I just realized this. I will try to fix the build before merging for safety.
https://github.com/travis-ci/travis-ci/issues/6120

miloyip:
Hi, I tried on OSX and seems the compiler version of clang is different. I tried to fixed with this:
~~~
--- a/test/unittest/CMakeLists.txt
+++ b/test/unittest/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(CheckCXXCompilerFlag)
+
 set(UNITTEST_SOURCES
        allocatorstest.cpp
     bigintegertest.cpp
@@ -41,7 +43,10 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     # If the user is running a newer version of Clang that includes the
     # -Wdouble-promotion, we will ignore that warning.
     if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 3.7)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-double-promotion")
+        CHECK_CXX_COMPILER_FLAG("-Wno-double-promotion" HAS_NO_DOUBLE_PROMOTION)
+        if (HAS_NO_DOUBLE_PROMOTION)
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-double-promotion")
+        endif()
     endif()
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     # Force to always compile with /W4
~~~

Besides, there are problems in MSVC compilation in appveyor. Can you check if they can be fixed?

miloyip:
I merged #654 and fixed some compiling issues with that.
Hope this PR can be updated as well. Thanks.

efidler:
I just pushed an update. Sorry for the slow progress; I'm travelling.

efidler:
ok, build passed

miloyip:
Thank you very much.

include/rapidjson/internal/dtoa.h
Thrive2021: I want to ask a question here. Since the array kPow10 is of length 10, why not check "index < 10" here? If index is 9, then it is still safe to access kPow10[9]. @efidler

PR #654, opened 2016-06-08
switch Clang from broken Ubuntu PPA to upstream binary release

efidler:
This should fix the Travis clang builds

https://github.com/travis-ci/travis-ci/issues/6120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,12 @@
+sudo: required
+dist: precise
+
 language: cpp
-sudo: false
 cache:
   - ccache
 
-addons:
-  apt:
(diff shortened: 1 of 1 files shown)


miloyip:
This seems not working in Travis. Any idea?

efidler:
Yeah, it's not working yet. The current issue is that the 32-bit builds need 32-bit libc++, which isn't in the upstream binary x86_64 release of clang+llvm. What do you think of just using the clang 3.5 included in Ubuntu trusty?

miloyip:
If clang 3.5 can compile properly, I think we can temporarily fallback to that.

PR #673, opened 2016-06-30
Fix buffer overrun using PutN (closes #672)

### LINKED ISSUE #672: Assert when outputting JSON using PrettyWriter
Using PrettyWriter to output JSON I encounter an assert when PutN is used to indent the output.

I hit the same assert with both Visual C++ 2008 and GCC 4.7.

Assertion failed: stackTop_ + sizeof(T) * count <= stackEnd_, file d:\p4\thirdpartyexports\rapidjson\export\1.0\1.0.3a2\includes\rapidjson\internal/stack.h, line 129

It appears to be an issue in PutN in stream.h. It instantiates templates differently for PutReserve and PutUnsafe. This causes it to use use PutReserve from stream.h (which does nothing), then uses PutUnsafe from stringbuffer.h (which assumes PutReserve allocated space).

Contents of stream.h PutN which causes assertion:
```
   //! Put N copies of a character to a stream.
   template<typename Stream, typename Ch>
   inline void PutN(Stream& stream, Ch c, size_t n) {
       PutReserve<Stream>(stream, n);
       for (size_t i = 0; i < n; i++)
           PutUnsafe(stream, c);
   }
```

Removing the template argument from PutReserve seems to resolve the issue, as it now uses PutReserve from stringbuffer.h.

Revised stream.h PutN which does not cause assertion:
```
   //! Put N copies of a character to a stream.
   template<typename Stream, typename Ch>
   inline void PutN(Stream& stream, Ch c, size_t n) {
       PutReserve(stream, n);
       for (size_t i = 0; i < n; i++)
           PutUnsafe(stream, c);
   }
```

sjasonsmith:
Fix inconsistent calling of template functions in PutN in stream.h. When
used with a GenericStringBuffer<<UTF8>, MemoryPoolAllocator>, PutN would call
PutReserve from stream.h, and PutUnsafe from stringbuffer.h. This
resulted in bytes being added to the buffer without allocating space.

This was not an issue when used with the default memory allocator,
because in this case the specialized PutN is used from stringbuffer.h.

### DIFF
--- a/include/rapidjson/stream.h
+++ b/include/rapidjson/stream.h
@@ -95,7 +95,7 @@ inline void PutUnsafe(Stream& stream, typename Stream::Ch c) {
 //! Put N copies of a character to a stream.
 template<typename Stream, typename Ch>
 inline void PutN(Stream& stream, Ch c, size_t n) {
-    PutReserve<Stream>(stream, n);
+    PutReserve(stream, n);
     for (size_t i = 0; i < n; i++)
         PutUnsafe(stream, c);
 }
--- a/test/unittest/stringbuffertest.cpp
+++ b/test/unittest/stringbuffertest.cpp
@@ -37,6 +37,13 @@ TEST(StringBuffer, Put) {
     EXPECT_STREQ("A", buffer.GetString());
 }
 
+TEST(StringBuffer, PutN_Issue672) {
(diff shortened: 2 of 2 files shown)


miloyip:
Thank you

PR #674, opened 2016-07-01
Make GenericSchemaDocument constructor explicit

sjasonsmith:
Prior to this change, a user could incorrectly pass a Document object to
SchemaValidator. This would implicitly construct a SchemaDocument, which
would then be destructed before the validator was used. This caused
unpredictable results including memory corruption and program crashes.

Consider the following snippet, which compiles cleanly but does not behave as expected, because the wrong variable was passed to the SchemaValidator constructor:

```
    Document sd;
    sd.Parse("{}");
    SchemaDocument schema(sd);
    SchemaValidator validator(sd);

    Document testDoc;
    testDoc.Parse("10");
    testDoc.Accept(validator);
```


miloyip:
Thanks

PR #676, opened 2016-07-06
Check for VS 2015: it doesn`t support stdint.h

aalex:
Visual Studio 2015 doesn't support stdint.h yet, so we should include msinttypes/stdint.h instead of stdint.h

### DIFF
--- a/include/rapidjson/rapidjson.h
+++ b/include/rapidjson/rapidjson.h
@@ -159,7 +159,7 @@
 */
 #ifndef RAPIDJSON_NO_INT64DEFINE
 //!@cond RAPIDJSON_HIDDEN_FROM_DOXYGEN
-#if defined(_MSC_VER) && (_MSC_VER < 1800)	// Visual Studio 2013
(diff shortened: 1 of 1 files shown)


pah:
Why is this needed in your environment? What error do you see otherwise?

The [Appveyor build](https://ci.appveyor.com/project/miloyip/rapidjson/branch/master) is successful on MSVC 2015 for the current `master`, without using the local `stdint.h` implementation.

aalex:
It seems like I can't reproduce the error right now. 
It's documented here: http://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio

pah:
The stackoverflow question is pretty old (2008).  In the answers, it's also documented, that Visual Studio supports `<stdint.h>` (to some extent) since 2010.

miloyip:
[Appveyor](https://ci.appveyor.com/project/miloyip/rapidjson/branch/master) shows that RapidJSON can be successfully built under VC2010 to VC2015.
So I close this now.

PR #681, opened 2016-07-11
fix rapidjson::value::Get<std::string>() may returns wrong data

### LINKED ISSUE #680: rapidjson::Value::Get<std::string>() may returns wrong data
`

    const char str[] = "value nonvalue";
    rapidjson::Value value(rapidjson::StringRef(str, 5));
    cout << value.Get<std::string>() << endl;
`

output
value nonvalue

expect output
value

niukuo:
fix #680


miloyip:
Thanks

PR #690, opened 2016-07-21
optimize FindMember  when use std::string

fuzhufang:
if define RAPIDJSON_HAS_STDSTRING, FindMember use std::string, but it also use internal::StrLen to get the string lengtht,
when it call FindMember(StringRef(name)).
Now use GenericValue construct it, then can use the std::string.size.
now it will be faster.


miloyip:
Thanks

PR #723, opened 2016-08-29
fix wrong length in remote schema

niukuo:
also hope to fix in v1.1.0


miloyip:
Thank you

miloyip:
Fixed #726

niukuo:
@miloyip will this fix release in v1.1.0?

miloyip:
Cannot change existing release. Need to make a new minor release.

PR #742, opened 2016-09-21
Fix some usage of rvalue references

0x1997:
`std::move` is missing in a few places.


pah:
These explicit ` std::move`s are not needed due to RapidJSON's default use of [move semantics][1].  They are even harmful, because they lead to an infinite recursion. ;-)

[1]: http://rapidjson.org/md_doc_tutorial.html#MoveSemantics

0x1997:
@pah I see. Thanks.

PR #748, opened 2016-09-25
Fix for winmindef.h defining min/max macros

sfinktah:
winmindef.h defines macros for `min` and `max` which cause several lines in documents.h to be expanded incorrectly whilst only presenting warnings that min and max lacked sufficient arguments.

See: http://stackoverflow.com/questions/22744262/cant-call-stdmax-because-minwindef-h-defines-max for more information.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -29,6 +29,14 @@ RAPIDJSON_DIAG_PUSH
 #ifdef _MSC_VER
 RAPIDJSON_DIAG_OFF(4127) // conditional expression is constant
 RAPIDJSON_DIAG_OFF(4244) // conversion from kXxxFlags to 'uint16_t', possible loss of data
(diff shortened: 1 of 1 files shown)


miloyip:
These macros are not defined by Visual C++ actually.
They should have been defined by WinDef.h.
You may defined NOMINMAX macro to prevent defining such macros.
It is not safe to just undef them in document.h as there is no way to restore them.

chwarr:
There's another approach that works across compilers. Adding an extra set of parenthesis. Would you prefer that instead of pushing/popping macro definitions? I can code it up and submit a PR.

```
static_cast<double>(std::numeric_limits<uint64_t>::max())
```
becomes
```
static_cast<double>((std::numeric_limits<uint64_t>::max)())
```

If you want me to code this up, there are some uses of min/max in the tests as well. Shall I guard them as well, or just the header files?

miloyip:
@chwarr Can you provide a PR for this? I think all source code is better. Thanks

chwarr:
https://github.com/miloyip/rapidjson/pull/1057 opened with the paren guard for min/max.

PR #756, opened 2016-09-30
Remove arch specific compiler flags.

AdamMajer:
While we are not really building anything that is distributed,
these compiler flags don't exist on all architectures.

On PPC this dies with.
  c++: error: unrecognized command line option '-march=native'; did you mean '-mcpu=native'?


miloyip:
Can you find out before which gcc version this is not supported? And then just disable it when it is not supported? (I don't know if cmake can automatically detect this.)
`-march=native` is needed to test most optimized code.

AdamMajer:
On 10/01/2016 09:49 AM, Milo Yip wrote:
> Can you find out before which gcc version this is not supported? And
> then just disable it when it is not supported? (I don't know if cmake
> can automatically detect this.)
> |-march=native| is needed to test most optimized code.

This applies to architectures, not compiler versions. I think it is
failing on PPC64 and some other that do not have that flag. I'll look
for a better way of disabling this on unsupported arches.

Cheers,
Adam

PR #760, opened 2016-10-04
FIx typo in documentation

icechen1:
The use of the vertical bar seems to break the rendering of the table in the documentation here:
http://rapidjson.org/md_doc_schema.html

Seems like we can fix it by using html encoding for the vertical bars as described in this post:
http://stackoverflow.com/questions/17319940/how-to-escape-a-pipe-char-in-a-code-statement-in-a-markdown-table/17320389#17320389


miloyip:
Thank you very much.

PR #789, opened 2016-11-04
Add optional length parameter to GenericStringStream constructor

masseman:
In order to be able to reference sub-strings (segments that are not zero-terminated) in `GenericStringStream,` a length parameter is necessary.

If the length parameter is not used, the compiler will optimize away the comparisons added in `Peek()` and `Tell()`, so this modification should not have any impact on performance.

### DIFF
--- a/include/rapidjson/stream.h
+++ b/include/rapidjson/stream.h
@@ -110,10 +110,11 @@ template <typename Encoding>
 struct GenericStringStream {
     typedef typename Encoding::Ch Ch;
 
-    GenericStringStream(const Ch *src) : src_(src), head_(src) {}
+    GenericStringStream(const Ch *src) : src_(src), head_(src), end_(0) {}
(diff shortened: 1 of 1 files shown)


stervo4ka:
> If the length parameter is not used, the compiler will optimize away the comparisons added
That's not true in general. Compiler might optimize the comparisons in simple cases, but consider this option:
```
void f(StringStream &stream) {
    // Do something.
}
```

In this case compiler cannot know whether the length is used.
Here is a proof https://godbolt.org/g/qMlFyH

miloyip:
There is already a [MemoryStream](https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/memorystream.h) which can be used, though lack of documentation.

masseman:
@miloyip Thanks, I will use that instead!
@stervo4ka Good point.

PR #827, opened 2017-01-18
Treat signed-unsigned conversions as errors.

lichray:
It's bad.  Turning on this warning gives us more chances to capture bad codegen (like the one explained here: https://youtu.be/yG1OZ69H_-o?t=39m17s).

Note on `CMakeLists.txt` changes: Clang's `-Wconversion` option turns on `-Wsign-conversion`; gcc doesn't.


include/rapidjson/writer.h
```diff
@@ -352,7 +352,7 @@ class Writer {
         char* end = internal::dtoa(d, buffer, maxDecimalPlaces_);
         PutReserve(*os_, static_cast<size_t>(end - buffer));
         for (char* p = buffer; p != end; ++p)
-            PutUnsafe(*os_, static_cast<typename TargetEncoding::Ch>(*p));
+            PutUnsafe(*os_, static_cast<typename OutputStream::Ch>(*p));
```
lichray: This change looks suspicious.  At least, it means rapidjson's coverage is low in Unicode area.  I think my change is right and the rest (in this file, which are using `TargetEncoding::Ch`) are not though.

miloyip: Yes. From the definition of `PutUnsafe()`:

~~~cpp
template<typename Stream>
inline void PutUnsafe(Stream& stream, typename Stream::Ch c) {
    stream.Put(c);
~~~

It should use the type in output stream. But I cannot think of any situation that these two character types are different.

include/rapidjson/internal/dtoa.h
```diff
@@ -41,7 +41,7 @@ inline void GrisuRound(char* buffer, int len, uint64_t delta, uint64_t rest, uin
     }
 }
 
-inline unsigned CountDecimalDigit32(uint32_t n) {
+inline int CountDecimalDigit32(uint32_t n) {
```
miloyip: I am not sure why these are needed to be changed to signed. Number of digits must be >= 0.

lichray: The point is that signed integer types should almost always be the choice whenever an arithmetic operation (+, -, *, /, etc) applys on that object, regardless whether the range of the value may be negative or not, because the wrapping behavior of unsigned integer types are very hard for the tools (such as UBsan) to reason your program, and often causes the compilers to produce downgraded code.  The YouTube video I pasted above shows a typical problem: using 32-bit unsigned integer in array offset on 64-bit architecture causes the compiler to "simulate" the wrapping behavior to ensure the address not overflowing the 64-bit address space, by changing the offset to 32-bit signed integer then the program is transformed into using only one 64-bit register because overflowing signed integer type itself is UB.  In summary, using `int` as raw array index is just fine, using `unsigned` may be bad for architectures with longer word.

Going back to the return value here, it's later being used to access array, very similar to the example above, so there is nothing wrong to changed it to signed.

include/rapidjson/writer.h
lichray: The warning triggers when `Steam::Ch` is `wchar_t` (UTF16 iirc) and `TargetEncoding` is `UTF8<>`.

miloyip:
I am convinced. Thank you. I try to fix the character type issue.

PR #842, opened 2017-02-03
Token-by-token pull parsing

StilesCrisis:
Refactored the iterative parser so that users can parse a single JSON
element at a time (invoking the handler one time) and then return
control to the calling code. Call IterativeParseInit to start, and then
call IterativeParseNext to retrieve one JSON element. Use
IterativeParseComplete to check for JSON document completion.


miloyip:
Thank you. Please update user guide for this new feature if possible.

hidva:
https://github.com/Tencent/rapidjson/issues/1714

@StilesCrisis Could you help to look at it?

Issue #845, opened 2017-02-06
internal::StrLen should be specialized for char with std::strlen


jcelerier:
The compiler seems to be able to use its builtins in this case, which can precompute the size at compile time (unlike `internal::StrLen`) : 

* GCC : https://godbolt.org/g/pslUOH
* MSVC : https://godbolt.org/g/SZsbGZ
* Clang is able to optimize it : https://godbolt.org/g/kqXBJ8

This could be important when using `write.String("someString")` which must be a quite common case.

StilesCrisis:
Sounds good to me. I've added this change as pull request #872.

StilesCrisis:
I just added a further template trick to allow quoted strings to know their length without relying on strlen at all. Requires some SFINAE trickery but should be a win overall.

harrywong:
Hi @StilesCrisis , I just face a problem about this pr, see #889 .
In brief, when you use a char array in a struct to store text, template auto-detect size will break the behavior the old rapidjson do. Because it will not detect the real string size at all. Maybe we can distinguish the C-style string and the char array field in a struct.

StilesCrisis:
Removed template trick.

PR #872, opened 2017-02-28
Use native strlen

StilesCrisis:
Compilers generally provide a much smarter strlen than ours. Resolves Issue #845


StilesCrisis:
Valgrind is complaining about __wcslen_sse2 on some 32-bit configurations overrunning the buffer, but that is not a bug in rapidjson (and is very likely to be known/not a bug). Don't know how you would prefer to handle that.

StilesCrisis:
Ah, known bug in valgrind. https://bugs.kde.org/show_bug.cgi?id=307828

StilesCrisis:
It doesn't look like Travis CI ran at all...?

miloyip:
May due to [Issues related to the S3 outage in AWS.](https://www.traviscistatus.com/incidents/hmwq9yy5dh9d)

StilesCrisis:
Sure enough, looks like Travis is working now.

StilesCrisis:
Whew, at long last, it looks like I might have placated Valgrind!

include/rapidjson/prettywriter.h
```diff
@@ -184,8 +186,16 @@ class PrettyWriter : public Writer<OutputStream, SourceEncoding, TargetEncoding,
...
-    bool String(const Ch* str) { return String(str, internal::StrLen(str)); }
-    bool Key(const Ch* str) { return Key(str, internal::StrLen(str)); }
+    template <typename T>
+    bool String(const T* const& str, RAPIDJSON_ENABLEIF((internal::IsSame<Ch, T>))) { return String(str, internal::StrLen(str)); }
+    template <typename T>
+    bool Key(const T* const& str, RAPIDJSON_ENABLEIF((internal::IsSame<Ch, T>))) { return Key(str, internal::StrLen(str)); }
+    
+    //! The compiler can give us the length of quoted strings for free.
+    template <typename T, size_t N>
+    bool String(const T (&str)[N], RAPIDJSON_ENABLEIF((internal::IsSame<Ch, T>))) { return String(str, N-1); }
```
stervo4ka: This overload makes string "abc" in the following case:
```
char abcd[] = {'a', 'b', 'c', 'd'};
writer.String(abcd);
```
This behaviour is unexpected, and that's why I generally avoid this trick in favor of strlen.

StilesCrisis: That code is illegal and would not have worked in rapidjson before or after my patch. Your string is not null terminated. The behavior without my patch will be undefined and definitely not what you want--you won't get 3, but you may well get a crash and will certainly get garbage data in your JSON. After my patch it's wrong in a slightly different way but that's irrelevant AFAICS. It's invalid code.

StilesCrisis: Missed a word... you won't get 3 *characters*.

stervo4ka: With your patch it's perfectly legal (because you add an overload specifically for arrays of chars), but works not as expected, and this is a problem.

Also I have to say that lots of tricky overloads with sfinae magic in public API complicates things for users.

StilesCrisis: In what way does it work unexpectedly?

If you expected your example code above to work in an expected way, that's invalid code invoking undefined behavior, and does not work before or after my patch.

stervo4ka: > In what way does it work unexpectedly?
In a way that it truncates one element from the array.

> that's invalid code invoking undefined behavior
As I already pointed out, it's not true. With your patch this code doesn't invoke UB (or elaborate how it does) and it's not invalid in any way (or elaborate how it is).
If you want users to assume that there's only an overload for c-strings and keep this new overload an implementation detail then it's not going to work because this new overload for arrays will be part of the public API and will allow users to write code as above.

StilesCrisis: I've added an assertion that the passed-in char array must be null-terminated, and a comment explaining this. I think this should be sufficient.

StilesCrisis: That being said, this change is breaking MSVC x64 anyway; if I can't figure out how to solve that, this all may end up being irrelevant :) Maybe I should split this into two separate pull requests.

StilesCrisis:
Whew! That got it working again.

miloyip:
Thank you for your continuous effort!

PR #881, opened 2017-03-08
Fix -Werror=effc++ errors with GNU 6.3.1

TedLyngmo:
Fix "'MyHandler::type’ should be initialized in the member
initialization list [-Werror=effc++]" errors.

https://github.com/miloyip/rapidjson/issues/874

### DIFF
--- a/example/simplepullreader/simplepullreader.cpp
+++ b/example/simplepullreader/simplepullreader.cpp
@@ -16,6 +16,14 @@ struct MyHandler {
     const char* type;
     std::string data;
     
+    MyHandler() : type(), data() { Null(); }
+    MyHandler(const MyHandler& cpy) : type(cpy.type),data(cpy.data) {}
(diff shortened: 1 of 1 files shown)


StilesCrisis:
I don't think the copy constructor or assignment operator add value. The handler object wasn't meant to be copied and the example is meant to be simple, not exercise every corner of C++. If you must include something, can you at least make them private/body-less at the bottom of the handler class? i.e.:

private:
    MyHandler(const MyHandler& noCopyConstruction);
    MyHandler& operator=(const MyHandler& noAssignment);

Also, there's no need to call Null() in the constructor, that only makes things confusing. It makes it look like the JSON stream begins with a null object.

StilesCrisis:
Thank you 👍

TedLyngmo:
I agree. I added the minimum I could  of to make g++ -Werror=effc++ happy and when doing it, I tried making the methods working too, but it's better to hide them like you propose. :-)

Issue #889, opened 2017-03-13
Issues about auto-sized by temple


harrywong:
I found an issue when updating rapidjson to the latest version. I found there's some issue about the auto-sized template updates.

I have a struct has a char[16] field. When the text is 'abcdefg\0', in the previous rapdjson can handle it using the strlen() method, while will wirte abcdefg in the json. In the new version, it will write all the \0 to the end of string, which is not the same behavior as the old version.

One workaround is to call strlen myself in the program.

StilesCrisis:
Hmm, OK. I'll revert that change.

StilesCrisis:
OK, Pull Request #890 fixes the issue by reverting the change, and includes a unit test which confirms your findings. Sorry about this!

harrywong:
Thanks very much!

StilesCrisis:
Just realized I forgot to address the same bug in PrettyWriter. Added PR #897 to handle this.

PR #890, opened 2017-03-13
Issue 889

StilesCrisis:
This fixes Issue #889 and includes a unit test detecting the issue.


stervo4ka:
omg why did you even argue

miloyip:
What do you mean? I think reverting first is the safe way to go. Isn't this your intention?

PR #896, opened 2017-03-15
Writer flush api

StilesCrisis:
Add a Flush() API to Writer for convenience.


StilesCrisis:
This also includes PR #891 (to make Clang happy).

PR #907, opened 2017-03-22
Fixed bug on space hexadecimal encoding

almavi:
Hi,

There was a bug regarding special characters in strings. Space is 32 (0x20) and code wants to check for 31 and below. However, 31 in hexadecimal is 0x1F, no 0x19.

By the way, I saw some suspicious 0x19999... and 0x9999... checks for i64 overflow. These are likely 0x1FFF and 0xFFF... but I haven't fixed them.

Alejandro


miloyip:
Can u add unit tests for this?

almavi:
I added tests for the reader and writer. These should work now, they should have failed in the past (the 0x1C encoding wasn't covered before).

almavi:
The test I wrote expects ascii characters to fail as "kParseErrorStringInvalidEncoding", but they are set as "kParseErrorStringEscapeInvalid". Line 958 of reader.h is

`RAPIDJSON_PARSE_ERROR(kParseErrorStringEscapeInvalid, is.Tell());`

and I think it should be

`RAPIDJSON_PARSE_ERROR(kParseErrorStringInvalidEncoding, is.Tell());`

but I will let you decide: shall I fix the test or reader.h?

miloyip:
Yes, I think it should be `kParseErrorStringInvalidEncoding`.
Please fix that also. Thank you very much.

PR #930, opened 2017-04-14
RAPIDJSON_STATIC_ASSERT: use C++11 static_assert, if available

pah:
This PR adds support for using C++11 `static_assert` in the `RAPIDJSON_STATIC_ASSERT` macro when it's available.  This has been suggested by @lichray in #924.

For readability, the C++11 implementation is added _before_ the C++03 implementation thus avoiding further conditional nesting.

The token concatenation macro `RAPIDJSON_JOIN`  is used in the diagnostics macros, so the related defines are moved outside the C++03 `RAPIDJSON_STATIC_ASSERT` implementation block.

### DIFF
--- a/include/rapidjson/rapidjson.h
+++ b/include/rapidjson/rapidjson.h
@@ -49,6 +49,11 @@
 // token stringification
 #define RAPIDJSON_STRINGIFY(x) RAPIDJSON_DO_STRINGIFY(x)
 #define RAPIDJSON_DO_STRINGIFY(x) #x
+
+// token concatenation
+#define RAPIDJSON_JOIN(X, Y) RAPIDJSON_DO_JOIN(X, Y)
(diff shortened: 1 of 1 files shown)


miloyip:
Thanks

enchantner:
I believe it's this patch which leads to "static_assert declarations are incompatible with C++98" error when I try to build it on Mac OS. "-Weverything" introduces "-Wc++98-compat" which doesn't work in this case.

pah:
See [this post](https://softwareengineering.stackexchange.com/questions/122608/clang-warning-flags-for-objective-c-development/124574#124574) from Chandler Carruth (Clang developer at Google) on ` -Weverything`:

> `-Weverything`: This is an insane group that literally enables every warning in Clang. Don't use this on your code. It is intended strictly for Clang developers or for exploring _what warnings exist_.

miloyip:
I also reproduced this problem after upgrading Xcode. I goto remove `-Weverything`.

Issue #849, opened 2017-02-10
Inconsistent reporting of numbers that are too big to be stored in a double


kahatlen:
If a JSON document contains a floating-point number that is too big to be stored in a double, RapidJSON sometimes raises an error that says "Number too big to be stored in double", and other times it silently accepts it and returns positive or negative infinity. It would be better if it consistently reported an error when it encountered a number that was too big.

Take this test program:

```c++
#include <iostream>
#include "rapidjson/error/en.h"
#include "rapidjson/error/error.h"
#include <rapidjson/reader.h>
#include <rapidjson/stream.h>

using namespace rapidjson;

class Handler : public BaseReaderHandler<UTF8<> > {
public:
  bool Double(double d) {
    std::cout << d;
    return true;
  }
};

void test(const char* json) {
  std::cout << json << " -> ";
  Reader reader;
  StringStream stream(json);
  Handler handler;
  if (!reader.Parse(stream, handler)) {
    std::cout << GetParseError_En(reader.GetParseErrorCode());
  }
  std::cout << std::endl;
}

int main() {
  test("3.14");
  test("1e308");
  test("1.7e308");
  test("1.8e308");
  test("5e308");
  test("1e309");
  test("1.0e310");
  test("1.00e310");
  test("-1.7e308");
  test("-1.8e308");
  test("-1e309");
}
```

With g++ 6.3.0 on Debian 9 (amd64), I see the following output:

```
3.14 -> 3.14
1e308 -> 1e+308
1.7e308 -> 1.7e+308
1.8e308 -> inf
5e308 -> inf
1e309 -> Number too big to be stored in double.
1.0e310 -> Number too big to be stored in double.
1.00e310 -> inf
-1.7e308 -> -1.7e+308
-1.8e308 -> -inf
-1e309 -> Number too big to be stored in double.
```

It would be more consistent if those cases that currently say "inf" or "-inf" said "Number too big" instead.

StilesCrisis:
I tested this with the Full Precision flag but it actually made matters worse. The "inf" entries became garbage:

nan
-8.6927218887537693e-309
-3.0943460473825783e-307
nan

jveazey:
What about using strtod or std::stod for converting value? The string could still be tested, as it currently is, to see if it contains the appropriate characters and format, but the final conversion could be done by a standard function.  If it fails, then errno will be set to HUGE_VAL and kParseErrorNumberTooBig can be returned.

PR #935, opened 2017-04-27
Fix for issue #849, standardise error message.

kAngelProjects:
For issue #849 - Added following code to strtod.h to make error message more consistent. 

`inline double FastPath(double significand, int exp) {
	if (exp < -308)
		return 0.0;
	else if (exp >= 0)
	{
		if (std::isinf(significand * internal::Pow10(exp))) {
			std::cout << "Number too big to be stored in double. ";
			return INFINITY;
		}
		else {
			return significand * internal::Pow10(exp);
		}
	}
	else {
		if (std::isinf(significand / internal::Pow10(-exp))) {
			std::cout << "Number too big to be stored in double. ";
			return -INFINITY;
		}
		else {
			return significand / internal::Pow10(-exp);
		}
		
	}
}`

### DIFF
--- a/include/rapidjson/internal/strtod.h
+++ b/include/rapidjson/internal/strtod.h
@@ -24,12 +24,28 @@ RAPIDJSON_NAMESPACE_BEGIN
 namespace internal {
 
 inline double FastPath(double significand, int exp) {
-    if (exp < -308)
-        return 0.0;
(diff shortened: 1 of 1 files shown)


OlafvdSpek:
std::cout?

PR #951, opened 2017-05-14
Addition of convenience methods to write pairs

hwiesmann:
Writing objects pairs is made easier with this commit.

PairNull is needed to prevent method calling clashes between

Pair(const Ch* key, SizeType, bool = false); and
Pair(const Ch* key, SizeType, boo, bool =false);

and similar methods.


hwiesmann:
Sorry, invalid methods called in case RAPIDJSON_HAS_STDSTRING is defined.

PR #952, opened 2017-05-14
Addition of convenience methods for writing pairs

hwiesmann:
These convenience methods make writing object pairs easier and better to read, IMHO.


miloyip:
I suggest not adding this kind of helper functions into the API. You may add them in your project as free functions.

PR #961, opened 2017-05-17
Add kParsePartialFractionFlag

gergondet:
This flag relaxes the requirements on double parsing so that fractions
can begin with a dot or have a dot without fraction digits.

This is related to some of the points in #36

I am not too keen on the flag name but I don't have a more inspired proposal.

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -155,6 +155,7 @@ enum ParseFlag {
     kParseNumbersAsStringsFlag = 64,    //!< Parse all numbers (ints/doubles) as strings.
     kParseTrailingCommasFlag = 128, //!< Allow trailing commas at the end of objects and arrays.
     kParseNanAndInfFlag = 256,      //!< Allow parsing NaN, Inf, Infinity, -Inf and -Infinity as doubles.
+    kParsePartialFractionFlag = 512,//!< Allow parsing fractions that begin or end with a dot.
     kParseDefaultFlags = RAPIDJSON_PARSE_DEFAULT_FLAGS  //!< Default parse flags. Can be customized by defining RAPIDJSON_PARSE_DEFAULT_FLAGS
 };
 
@@ -1529,7 +1530,8 @@ private:
                 RAPIDJSON_PARSE_ERROR(kParseErrorValueInvalid, s.Tell());
             }
         }
-        else
+        // Continue parsing if it looks like a partial fraction
+        else if (!((parseFlags & kParsePartialFractionFlag) && s.Peek() == '.'))
             RAPIDJSON_PARSE_ERROR(kParseErrorValueInvalid, s.Tell());
 
         // Parse 64bit int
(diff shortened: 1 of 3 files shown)


include/rapidjson/reader.h
```diff
@@ -1574,8 +1576,12 @@ class GenericReader {
         if (Consume(s, '.')) {
             decimalPosition = s.Length();
 
-            if (RAPIDJSON_UNLIKELY(!(s.Peek() >= '0' && s.Peek() <= '9')))
-                RAPIDJSON_PARSE_ERROR(kParseErrorNumberMissFraction, s.Tell());
+            if (RAPIDJSON_UNLIKELY(!(s.Peek() >= '0' && s.Peek() <= '9'))) {
+                if (!(parseFlags & kParsePartialFractionFlag))
+                    RAPIDJSON_PARSE_ERROR(kParseErrorNumberMissFraction, s.Tell());
+                else if (!(s.Peek() == ',' || s.Peek() == '}' || s.Peek() == ']' || s.Peek() == '\0' || s.Peek() == 'e' || s.Peek() == 'E'))
```
miloyip: I think this should not be needed here. This function should parse as many characters for a JSON number type as it could. It should not involve other characters such as `,}]\0`.

gergondet: Thanks for reviewing the code.

I think it is needed to accept a missing fraction.

In the standard case, we should outright reject the number if it does not have numbers following the `.`.

In the partial fraction case, we can accept the number to finish here (i.e. check for `,}]\0`) or to be followed by an exponent but it should not accept anything else.

Without this check, the parser works correctly witht incomplete fractions but the parser fails with `kParseErrorDocumentRootNotSingular` when parsing `.a123` rather than `kParseErrorValueInvalid` which might be a bit confusing. If you think this is ok, I'll hapilly remove that check.

miloyip: I think implementing this relaxed syntax just involve change of the underlying state machine.
For `123.a`, it should take `123.` and returns a double valued `123.0`.
For `.123a`, it should take `.123` and returns a double valued `.123`.
For `.a` it should raise a parse error because there must be at least a digit when it is prefixed by `.`.

gergondet: Ok, I've removed the extra checks then.

miloyip:
I just realized that the implementation relaxed the syntax such that even `.` and `.E10` are parsed as zero. I think this may be over-relaxed.

I checked [ECMAscript's syntax](https://www.ecma-international.org/ecma-262/8.0/index.html#prod-DecimalLiteral). The decimal number is either (with at least one digit before "." and without digits after), or (without digits before "." and with a least one digit after).

I think we may consider following this syntax, in order to make JSON's syntax relax to ECMAscript's , but not more.

kkaefer:
@miloyip what next actions need to happen to get this merged?

miloyip:
@kkaefer The discussion was suspended... Any idea to the above discussion?

PR #963, opened 2017-05-18
Escape solidus in writer

cerevra:
There are scenarios when we need strict rfc-compliance in this place
https://tools.ietf.org/html/rfc7159#section-7


miloyip:
> All Unicode characters may be placed within the
   quotation marks, except for the characters that must be escaped:
   **quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F)**.

Solidus does not need to be escaped.

cerevra:
Ok. You are right. But this RFC is inconsistent
>       char = unescaped /
>           escape (
>               %x22 /          ; "    quotation mark  U+0022
>               %x5C /          ; \    reverse solidus U+005C
>               %x2F /          ; /    solidus         U+002F
>               %x62 /          ; b    backspace       U+0008
>               %x66 /          ; f    form feed       U+000C
>               %x6E /          ; n    line feed       U+000A
>               %x72 /          ; r    carriage return U+000D
>               %x74 /          ; t    tab             U+0009
>               %x75 4HEXDIG )  ; uXXXX                U+XXXX

And rapidjson logic is inconsistent now too. Reader knows about escaped solidus, but writer ingnores it. https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/reader.h#L994

Consistency is good thing

miloyip:
There are multiple ways of encoding, but only one way of decoding.
I can just say that, this is by design. And RapidJSON is compliant with the standard.

Issue #962, opened 2017-05-18
Storing values for later use


kadst43:
I have perhaps a bit of an unusual use case...  I want to load in one or more files that contain user-created configuration data for my application.  I then want to parse the top layer or two of these files, pull out the values, and store them in my own data structure, for later use.  I'd rather not have to keep the original document around, if only because I don't know how many files I'll be loading (it's the user's prerogative to decide that).  So, my original design has been to load the file into a char*, parse it in situ, and then make deep copies of the values which I need to store for later use using the copy constructor.  

The problem that I'm having seems to be that the allocator is owned by the document, so when I clean up the document it also deletes the allocator, which deletes all of those deep copies that I made.  I tried using a separate allocator (not the one that belongs to the document), but that didn't appear to work (I think maybe it still uses the document's allocator for some things?  I'm not sure.).  I tried having a separate allocator, and then passing it into the document's constructor, but then the document takes ownership of it and cleans it up.  I tried using a CRT allocator,  but didn't have luck with that either.

So my question is, is there a clean way to do what I'm wanting to do?  I'd love to use the memory pool allocator if I can, I'd be willing to use the CRT allocator if I have to...

I can share code if I need to, but hopefully the issue is clear from what I wrote?  Since none of my code is right, I'm not sure how helpful it will be...

miloyip:
You may keep a document for each deep copy.
Allocator should work too. Can you show some code to show how it does not work?

kadst43:
Found the problem.  It wasn't the allocator or deleting the document (which I should have know), it was that I deleted the char* as well.  So this code is bad:

`    char* fileBuffer = /* allocate a char array on the heap, load the file into it */
    MemoryPoolAllocator<> alloc;
    Document* pDoc = new Document();
    pDoc->ParseInsitu(fileBuffer);
    // ... extract the values, make deep copies, store the copies
    delete pDoc;
    delete[] fileBuffer;   // <== THE PROBLEM!!!!!!
`

That begs another question I wanted to ask, though... what is the right way to clean up that char* in this situation, assuming that it's not created on the stack (which it can't be, because I don't know the file size until runtime)?  Do I just need to leak it?  Keep a pointer to it and delete it when I'm done with the stored values?  Or, alternately, should I just keep it around and then not bother to make the deep copies?

pah:
[...]

PR #964, opened 2017-05-18
GenericValue::CopyFrom: add option to force copying of strings

pah:
Copying the result of an in-situ parsing into another value/document currently requires that the original buffer - still holding the strings from the parsing, outlives the  destination object as well.

In order to obtain a "full" copy of a GenericValue, this commit adds an optional flag `copyConstStrings` to `CopyFrom`, which then forces to take a copy of all embedded strings in the source value.

This solves the problem discussed in #962.


miloyip:
Thanks

PR #967, opened 2017-05-25
Storage class is not first - encodings.h

TomaszNo:
If storage class is not specified as first in declaration then Intel C++ Compiler 2017 generates message:
message #82: storage class is not first


miloyip:
Thanks

PR #975, opened 2017-06-07
Added INTERFACE target RapidJSON

ksergey:
The patch gives ability to use RapidJSON as subproject maner:
```
add_subdirectory(deps/rapidjson)
...
add_executable(app main.cpp)
target_link_libraries(app RapidJSON)
```
No need additional include_directories or any other command in CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,9 @@ SET(CMAKE_INSTALL_DIR "${_CMAKE_INSTALL_DIR}" CACHE PATH "The directory cmake fi
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
+add_library(RapidJSON INTERFACE)
+target_include_directories(RapidJSON INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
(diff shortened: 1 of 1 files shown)


tambry:
You'll want to bump up the minimum version to 2.8.12, which added some of the `target_*` commands to the ancient CMake 2.8.

CMakeLists.txt
```diff
@@ -122,6 +122,9 @@ SET(CMAKE_INSTALL_DIR "${_CMAKE_INSTALL_DIR}" CACHE PATH "The directory cmake fi
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
+add_library(RapidJSON INTERFACE)
+target_include_directories(RapidJSON INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/include)
```
steveire: I recommend adding an `ALIAS` target too with `::` in the name. Perhaps `RapidJSON::RapidJSON`. This causes CMake to issue better diagnostics.

ClausKlein: see https://github.com/Tencent/rapidjson/pull/1810

PR #999, opened 2017-06-28
support jscript code's json

liuzilu19:
eg. { 1 : "a", b : "x"}


miloyip:
This would not be accepted as it changes the existing behavior.

StilesCrisis:
Using non-strings for object names is illegal JSON anyway. If the goal is to be a standards-compliant parser in default settings, we would need to gate this behavior behind a flag.

PR #1002, opened 2017-06-30
Tutorial: fix typos in examples and broken links

FluxLemur:
In the move example, the code uses `contacts` when the diagrams use
`contact` (no 's')

The code in the example:
```    Value contacts(kArrayType);
    // adding elements to contacts array.
    // ...
    o.AddMember("contacts", contacts, d.GetAllocator());  // deep clone contacts (may be with lots of allocations)
    // destruct contacts.
```

### DIFF
--- a/doc/diagram/move2.dot
+++ b/doc/diagram/move2.dot
@@ -18,7 +18,7 @@ digraph {
 
 		node [shape=Mrecord, style=filled, colorscheme=spectral7]
 
-		c1 [label="{contact:array|}", fillcolor=4]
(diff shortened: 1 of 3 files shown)


miloyip:
The paths in links are necessary for doxygen to generate HTML in http://rapidjson.org.

FluxLemur:
Ah I see. It's too bad it's broken in github, but certainly better to have a proper doxygen generated website.

I'll undo the path changes.

miloyip:
Thank you

Issue #989, opened 2017-06-16
Explicit operator bool for ParseResult


OlafvdSpek:
This currently compiles.. you might want to make operators bool explicit.
```
rapidjson::ParseResult a;
int b = a + 1;
```

PR #1008, opened 2017-07-09
ParseResult: improve bool conversion and add operator!=

pah:
As reported by @OlafvdSpek in #989, the implicit conversion to `bool` can lead to accidentally compiling expressions.  In addition to the mentioned arithmetic expressions, also using `operator!=` leads to wrong results:
```cpp
  ParseResult err = doc.Parse("{]");
  if (err != kParseErrorNone) {
    // missed! (err --> false, kParseErrorNone --> 0 --> equal!)
  }
```

Instead of using an `explicit operator bool` (requires C++11), this PR proposes using the safe-bool idiom.  For completeness, the various `operator!=` overloads are added as well.

### DIFF
--- a/include/rapidjson/error/error.h
+++ b/include/rapidjson/error/error.h
@@ -104,6 +104,8 @@ enum ParseErrorCode {
     \see GenericReader::Parse, GenericDocument::Parse
 */
 struct ParseResult {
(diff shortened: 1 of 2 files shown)


miloyip:
Thank you

Issue #817, opened 2017-01-04
GenericStringRef(NULL, 0)?


OlafvdSpek:
Apparently GenericStringRef requires the first argument to be non-NULL, I'm not sure it should.. if the size is 0. Why can't it support NULL here?

pah:
With the merge of #1010, I think we can close this issue.

Issue #226, opened 2015-01-13
Value of string type should not point to NULL


miloyip:
My colleague provided feedback that writing a document causes core dump. He find that it is due to a string pointed to NULL. A JSON string, according to the standard, cannot be JSON null type. An empty JSON string should be `""`.

Currently the constructor `GenericValue(kStringType)` and `SetString(0, 0)` should generate invalid string. Not sure if there is other cases as well.

I suggest that `GenericValue(kStringType)` should use `kShortStringFlag` for representing `""`. And other functions should generate an assertion failure.

Open for discussion.

miloyip:
Just checked, https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/document.h#L319 has already asserted:
~~~
    GenericStringRef(const CharType* str, SizeType len)
        : s(str), length(len) { RAPIDJSON_ASSERT(s != NULL); }
~~~

By the way, I found that another possible case is to use `MemberIterator` to change `name` into non-string type. This can be asserted in `Accept()`.

PR #1010, opened 2017-07-09
Improve handling of NULL strings

pah:
As discussed in #226 already, `Value::GetString()` shall never return a null string.  However, even with assertions enabled, the following code leads to a crash as the non-null assertion in ` GenericStringRef` comes too late:
```cpp
    const char* nullPtr = 0;
    Value v(StringRef(nullPtr)); // <-- segfault, calls StrLen with NULL
    v.SetString(nullPtr, allocator); // <-- segfault, calls StrLen with NULL
```

The fix for this is to embed the assert already before the `StrLen` call :
```cpp
  length(((RAPIDJSON_ASSERT(str != 0)), internal::StrLen(str)))
```
which requires changing the ` RAPIDJSON_ASSERT` override in ` unittest.h` to enable usage within expressions (f1ba61c).  All other code  paths are consistently routed through `GenericStringRef` to provide the required invariants.

As an extension suggested in #817, this PR allows passing null pointers with an explicit length of 0.  The implementation then still references an empty string internally to avoid returning `NULL` from `GetString()`.

Tests are added to cover the different cases.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -308,7 +308,7 @@ struct GenericStringRef {
      */
 #endif
     explicit GenericStringRef(const CharType* str)
-        : s(str), length(internal::StrLen(str)){ RAPIDJSON_ASSERT(s != 0); }
+        : s(str), length(((RAPIDJSON_ASSERT(str != 0)), internal::StrLen(str))) {}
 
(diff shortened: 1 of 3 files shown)


miloyip:
Thank you

PR #1021, opened 2017-07-14
MemoryPoolAllocator: fix reporting incorrect size.

TravisSmithBlz:
During 'AddChunk', the amount allocated is the requested capacty plus the
size of a ChunkHeader. The additional size of the ChunkHeader is not taken
into account when the size is reported in the 'Size' method. If the
difference between the buffer size and the allocated size is less than the
size of a ChunkHeader, the user will allocate off the heap each time and
not know to increase the buffer's size.

### DIFF
--- a/include/rapidjson/allocators.h
+++ b/include/rapidjson/allocators.h
@@ -168,7 +168,7 @@ public:
     size_t Size() const {
         size_t size = 0;
         for (ChunkHeader* c = chunkHead_; c != 0; c = c->next)
-            size += sizeof(ChunkHeader) + c->size;
+            size += RAPIDJSON_ALIGN(sizeof(ChunkHeader)) + c->size;
         return size;
     }
 
--- a/include/rapidjson/allocators.h
+++ b/include/rapidjson/allocators.h
@@ -166,9 +166,9 @@ public:
     /*! \return total used bytes.
     */
     size_t Size() const {
-        size_t size = 0;
+        size_t size = RAPIDJSON_ALIGN(sizeof(ChunkHeader));
         for (ChunkHeader* c = chunkHead_; c != 0; c = c->next)
-            size += RAPIDJSON_ALIGN(sizeof(ChunkHeader)) + c->size;
+            size += c->size;
         return size;
     }
 
--- a/include/rapidjson/allocators.h
+++ b/include/rapidjson/allocators.h
@@ -168,7 +168,7 @@ public:
     size_t Size() const {
         size_t size = 0;
         for (ChunkHeader* c = chunkHead_; c != 0; c = c->next)
-            size += RAPIDJSON_ALIGN(sizeof(ChunkHeader)) + c->size;
+            size += sizeof(ChunkHeader) + c->size;
         return size;
     }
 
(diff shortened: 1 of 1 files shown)


miloyip:
This may be a bit confused. Since `GetSize()` returns total number of bytes that the allocator allocated by `Malloc()`. This does not include the reserved (capcaity) bytes, or other overheads.

TravisSmithBlz:
There is a section in 'dom.md' called 'User Buffer'. In it, it says that a MemoryPoolAllocator can be used to avoid heap allocations. It will allocate off the heap when the supplied buffer is full.

 At the bottom it states: 
> "User can query the current memory consumption in bytes via MemoryPoolAllocator::Size(). And then user can determine a suitable size of user buffer."

In the definition of MemoryPoolAllocator, the constructor checks to ensure the passed buffer is at least large enough to hold a chunk header. This is because chunk headers take up space in the user supplied buffer. 

Line 129: 'allocators.h'
```
 RAPIDJSON_ASSERT(size > sizeof(ChunkHeader)); 
```

This chunk header is written into the beginning of the user buffer, and needs to be counted when the user is told how much space they need.

Later, in 'AddChunk', the MemoryPoolAllocator will call its base allocator's malloc because it has run out of memory in the buffer. 

Line 240: 'allocators.h'
```
baseAllocator_->Malloc(RAPIDJSON_ALIGN(sizeof(ChunkHeader)) + capacity))
```

The allocated memory includes the chunk headers. When the user wants to determine the required size of the buffer, the value returned by 'Size()' MUST include the chunk headers or it will be too small and heap allocations will continue.

include/rapidjson/allocators.h
```diff
@@ -168,7 +168,7 @@ class MemoryPoolAllocator {
     size_t Size() const {
         size_t size = 0;
         for (ChunkHeader* c = chunkHead_; c != 0; c = c->next)
-            size += sizeof(ChunkHeader) + c->size;
+            size += RAPIDJSON_ALIGN(sizeof(ChunkHeader)) + c->size;
```
TravisSmithBlz: I missed the alignment in the first commit.

include/rapidjson/allocators.h
```diff
@@ -178,9 +178,11 @@ class MemoryPoolAllocator {
             return NULL;
 
         size = RAPIDJSON_ALIGN(size);
-        if (chunkHead_ == 0 || chunkHead_->size + size > chunkHead_->capacity)
-            if (!AddChunk(chunk_capacity_ > size ? chunk_capacity_ : size))
+        if (chunkHead_ == 0 || chunkHead_->size + size > chunkHead_->capacity) {
+            size = chunk_capacity_ > size ? chunk_capacity_ : size;
```
TravisSmithBlz: This size will be assigned later, we need to ensure the correct value is used in the allocation case.

TravisSmithBlz: I'm reverting this, I misread the code.

TravisSmithBlz:
We only need to count the first chunk header since its stored in the memory buffer and the rest won't be needed if the buffer is large enough to hold all the data.

Issue #1024, opened 2017-07-21
SJSON support


bmuzzin:
SJSON is "Simplified JSON", which is a format used in some Autodesk products.
http://help.autodesk.com/view/Stingray/ENU/?guid=__stingray_help_managing_content_sjson_html

The SJSON format has the following differences from standard JSON:

- [ ] The equal sign = is used to define key-value pairs instead of the colon :.
- [ ] Quotes around string keys in key-value pairs are optional, unless you need the key to contain either spaces or the equal sign =.
- [ ] Commas are optional in object and array definitions.
- [x] Comments are allowed, using the C and C++ commenting styles:
// begins a single-line comment that causes everything to be ignored until the end of the line;
/* ... */ creates a multi-line comment that causes everything between the asterisks to be ignored.
- [ ] Each SJSON file is always interpreted as a definition for a single object. You can think of this as an implicit set of curly quotes { ... } that surround the contents of the file.

This has some overlap with a relaxed JSON #36 implementation. Comments are already implemented, and keys without quotes are also part of relaxed JSON.

https://github.com/Autodesk/sjson

1354092549:
SJSON is **quite different form** JSON and **JavaScript(ECMAScript)**,so I think it needs a separate library

But I think we need to provide a flag to support [JSON5](http://json5.org/) becasue now more and more JSON are written by humans and ***[JSON5](http://json5.org/) is a subset of JavaScript***

andrusha97:
@1354092549 What are the exact reasons to be compatible with javascript? I don't see how json5 is better to be in rapidjson than sjson. Both are "unofficial" extensions to json, and both could have their users if implemented in rapidjson.

1354092549:
@andrusha97 
SJSON is not compatible with JSON
JSON5 is a superset of JSON

If you just need to be simple,why not use YAML?
YAML is simpler than JSON/SJSON/JSON5

1354092549:
And it's very surprising that it is called JSON but isn't compatible with JavaScript

bmuzzin:
SJSON is very similar to JSON, although cannot be directly interpreted in Javascript. The intention of all these 'extensions' to JSON is to make it more human readable/editable, and many of the differences from JSON between them are similar. They could be implemented as small additional features on the rapidjson parser/writers, and combinations of flags can be used to implement different flavors of these extensions (see my PR #1025).

[...]

PR #1025, opened 2017-07-21
Issue #1024: SJSON support.

bmuzzin:
#1024 SJSON support.
http://help.autodesk.com/view/Stingray/ENU/?guid=__stingray_help_managing_content_sjson_html

Implements the following via Parse and Write flags:
- The equal sign is used to define key-value pairs instead of the colon :
- Quotes around strings in key-value pairs are optional
- Commas are optional in object and array definitions
- Each SJSON file is interpreted as a definition of a single object.

Allowing of comments was already implemented.

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -24,6 +24,7 @@
 #include "internal/stack.h"
 #include "internal/strtod.h"
 #include <limits>
+#include <locale>
 
 #if defined(RAPIDJSON_SIMD) && defined(_MSC_VER)
 #include <intrin.h>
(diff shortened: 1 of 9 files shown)


andrusha97:
There's asymmetry between kParseSJSONDefaultFlags and kWriteDefaultSJSONFlags. Shouldn't they be named alike?

bmuzzin:
Yes, makes sense. I've updated kWriteDefaultSJSONFlags to kWriteSJSONDefaultFlags.

stfx:
Good job! I like this especially for parsing humanly written JSON.

The only thing I noticed is that the pretty SJSON writer writes `key= value` instead of `key = value` (no spacing before equal) and that it also writes empty lines at start and end instead of omitting those originally top level lines.

Also the code does not follow rapidjsons zero overhead policy by introducing non-flag checks/variables which can't be compiled out unless I am mistaken.

Thanks for sharing your code!

include/rapidjson/reader.h
```diff
@@ -24,6 +24,7 @@
 #include "internal/stack.h"
 #include "internal/strtod.h"
 #include <limits>
+#include <locale>
```
stfx: not needed I believe

include/rapidjson/reader.h
```diff
@@ -155,7 +156,12 @@ enum ParseFlag {
     kParseNumbersAsStringsFlag = 64,    //!< Parse all numbers (ints/doubles) as strings.
     kParseTrailingCommasFlag = 128, //!< Allow trailing commas at the end of objects and arrays.
     kParseNanAndInfFlag = 256,      //!< Allow parsing NaN, Inf, Infinity, -Inf and -Infinity as doubles.
-    kParseDefaultFlags = RAPIDJSON_PARSE_DEFAULT_FLAGS  //!< Default parse flags. Can be customized by defining RAPIDJSON_PARSE_DEFAULT_FLAGS
+    kParseEqualReplaceColon = 512,  //!< sjson: = is used to define key-value pairs instead of the colon :.
+    kParseImplicitTopLevel = 1024,  //!< sjson: Each SJSON file is always interpreted as a definition for a single object.
+    kParseKeyQuotesOptional = 2048, //!< sjson: Quotes around string keys in key-value pairs are optional, unless you need the key to contain either spaces or the equal sign =.
+    kParseCommasOptional = 4096,    //!< sjson: Commas are optional in object and array definitions.
```
stfx: different order than new writer flags

include/rapidjson/reader.h
```diff
@@ -837,6 +867,9 @@ class GenericReader {
                     RAPIDJSON_PARSE_ERROR(kParseErrorTermination, is.Tell());
                 return;
             }
+            else if (parseFlags & kParseCommasOptional) {
+
+            }
             else
```
stfx: or maybe `else if (!(parseFlags & kParseCommasOptional))`

PR #1034, opened 2017-08-07
Allow Swap with ValueType

bluehero:


pah:
Hmmm, I'm not too happy about this change.  There's a  risk of inadvertently "swapping in" a value allocated from a different `Document` (and subsequently allocator).

If you really wanted to do this, you could already call swap the "other way round":
```cpp
  GenericDocument d;
  GenericValue v;
  v.SetArray().PushBack(true, d.GetAllocator());
  v.Swap( d ); // worked before already
```

Of course, the risk is only slightly less prominent when not bringing in the base class's overload into the `Document`'s interface.  Still, I have seen it as an advantage in the past.

test/unittest/documenttest.cpp
```diff
@@ -300,7 +300,14 @@ TEST(Document, Swap) {
     o.SetObject().AddMember("a", 1, a);
 
     // Swap between Document and Value
-    // d1.Swap(o); // doesn't compile
```
pah: This has been by design.

miloyip:
I am not sure if the asymmetric behavior is an issue. Should x.Swap(y) and y.Swap(x) have the same behavior?

Bu11etmagnet:
If x and y are the same type, most people would expect that and would be very surprised if it doesn't.

pah:
@miloyip said:
> I am not sure if the asymmetric behavior is an issue. Should x.Swap(y) and y.Swap(x) have the same behavior?
@Bu11etmagnet said:
> If x and y are the same type, most people would expect that and would be very surprised if it doesn't.

But here, we're talking about the case, where `v` and `d` have **different** types, i.e. `Value` vs. `Document`.  I would even say that `v.Swap(d)` could be disallowed without an explicit cast, because it is not obvious otherwise that you swap different types here.

miloyip:
Since `v.Swap(d)` is already in the codebase for awhile I think it would not be good time to remove it. I am thinking of making a new 2.0 API like `Document2` for fully C++11 or above support and fix many issues that the current 1.x API has.

PR #1048, opened 2017-08-29
Writer must escape solidus (/)

mkar71:
RFC 7159 lists the following escaped chars. solidus  is not being escaped currently.
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              **%x2F /          ; /    solidus         U+002F**
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009


FSMaxB:
No, solidus doesn't need to be escaped!

> The representation of strings is similar to conventions used in the C family of programming languages.  A string begins and ends with quotation marks.  All Unicode characters may be placed within the  quotation marks, except for the characters that must be escaped: quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).

It can be escaped, but it doesn't have to be.

FSMaxB:
Only the following need to be escaped (as cited in my last comment):
* quotation mark
* reverse solidus
* U+0000 trhough U+001F (control characters)

miloyip:
https://tools.ietf.org/html/rfc7159#section-7

> ... All Unicode characters may be placed within the
   quotation marks, except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).

So solidus is not required to be escaped. And I close this PR.

PR #1051, opened 2017-09-02
Specifies the endian of msvc ARM64 configuration.

gongminmin:
Windows ARM64 is also little endian.


miloyip:
Thank you

gongminmin:
You are so welcome.

PR #1052, opened 2017-09-02
Fix cross-compilation on certain toolchains

Crunkle:
On certain subsystems `CMAKE_SYSTEM_PROCESSOR` is unavailable due to cross-compilation via MinGW. This causes the expression to be evaluated incorrectly and throw the following message:

> CMake Error at CMakeLists.txt:53 (if):
>   if given arguments:
> 
>     "STREQUAL" "powerpc" "OR" "STREQUAL" "ppc64" "OR" "STREQUAL" "ppc64le"
> 
>   Unknown arguments specified

Nothing fancy. A very simple push to ensure that it is always evaluated string-against-string to avoid issues with less-verbose toolchain files which do not define this variable and leave it empty.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ if(CCACHE_FOUND)
 endif(CCACHE_FOUND)
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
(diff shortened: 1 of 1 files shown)


miloyip:
Thank you

PR #1055, opened 2017-09-06
Fix Windows doc build MSBuild error MSB6001

chwarr:
When using a MSBuild-based CMake generator like 'Visual Studio 15 2017
Win64', the doc build was failing with the error 'MSB6001: Invalid
command line switch for "cmd.exe". Illegal characters in path.'

This was due to the dependency on Doxyfile*, which wasn't expanded by
CMake.

The fix is to expand this glob in CMake before specifying the custom
command's dependencies.

Partial fix for https://github.com/miloyip/rapidjson/issues/622


miloyip:
Thanks

PR #1057, opened 2017-09-06
Guard against min/max being macros in a cross-compiler way

chwarr:
Guards against min/max being macros.

Sometimes, particularly when Microsoft's windows.h is included, min/max are defined as macros, interfering with use of std::numeric_limits::min() and the like.  To guard against this, the function name is wrapped in an extra set of parenthesis, which inhibits function-style macro expansion.

Two separate commits here: one for document.h and one for test code.


miloyip:
Thanks

Issue #1064, opened 2017-09-16
Schema: GetInvalidSchemaPointer returns empty pointer when a remote schema is violated


yurikhan:
A schema can include another schema from a different file, via the `$ref` keyword where the URI includes anything other than the fragment identifier.

When that included schema is violated, the schema pointer returned by `GetInvalidSchemaPointer()` is always the empty pointer.

A simplified test case: Validate `null` against the schema

```json
{"$ref": "foo#/integer"}
```

where `foo` refers (via a suitable `RemoteSchemaDocumentProvider`) to

```json
{"integer": {"type": "integer"}}
```

Expected: `GetInvalidSchemaPointer()` returns `Pointer("/integer")`.

Observed: `GetInvalidSchemaPointer()` returns `Pointer()`.

miloyip:
I am not sure how it should be handled.
For example, should it be `Pointer("foo#/integer")`?

yurikhan:
As far as I can tell, `Pointer`s can only express paths within a single JSON document, or, equivalently, the fragment part of an URI.

I am currently working towards a general solution for #619. I envision reporting violations in a JSON document and representing schema locations as string values in URI syntax, like `{"schemaRef": "foo#/integer"}`. I needed accurate in-schema paths for every case so I extracted this part as a separate improvement.

miloyip:
For creating new feature, keep in mind about backward compatibility. That is what RapidJSON always tried to do. Thanks.

yurikhan:
I strive to preserve the existing behavior, yes. However, if I discover surprising behavior that I would consider a bug, I report it as such. It’s then your call whether it’s worth changing.

yurikhan:
So, what’s your take? Should I preserve the existing behavior of `GetInvalidSchemaPointer()` to return an empty pointer when the actual subschema that is violated is in a different schema document?

It turns out to be not a big deal for #619. I can either change `GetInvalidSchemaPointer()` to return the accurate path, or leave it as is and just work around it in the new code.

Issue #619, opened 2016-04-28
There is no way to get which required field is missing on schema validation


rodrigostrauss:
I'm using the schema example from documentation. 

I didn't find any method to return the name of the missing field. I checked the code for Schema::EndObject and it doesn't save the name anywhere:

```    
   bool EndObject(Context& context, SizeType memberCount) const {
        if (hasRequired_)
            for (SizeType index = 0; index < propertyCount_; index++)
                if (properties_[index].required)
                    if (!context.propertyExist[index])
                        RAPIDJSON_INVALID_KEYWORD_RETURN(GetRequiredString());
``` 

The only thing I can get back from the SchemaValidator is the word "required".

miloyip:
Thank you for pointing out.
Yes, currently there was no additional information provided for invalid situation.
I will try to add error information for this.

rodrigostrauss:
I can write the fix myself if you give me some directions about the interface and where to store the data. I see that you don't just use some vector<string> to store stuff, so please point me on how to make a pull request you can actually accept.

PR #1068, opened 2017-09-22
Schema violation details

yurikhan:
I’d like to present this attempt at implementing detailed schema violation reporting. This covers #619 and other cases where the document is invalid but a simple triple (`GetInvalidDocumentPointer`, `GetInvalidSchemaPointer`, `GetInvalidSchemaKeyword`) is insufficient to spot the error.

All detected violations are collected into a JSON value and annotated with references into the invalid document and violated schema and useful details about each particular violation. The intent is to be as helpful to the reader as reasonable without putting too much burden on the validator.

Documentation for the error object format is provided and tests are updated to also check the error object.

I am open to suggestions on the API changes and on the error object structure, and prepared to invest the effort required to get this merged.

## Breaking API changes

* `SchemaDocument`: The main constructor now takes two more arguments — the string and length of the URI of this particular schema document. They are optional but come in front of the `IRemoteSchemaDocumentProvider` argument because if your application uses remote schemas it will probably want them to know their URIs.
  * An alternative solution would be to add a `SetURI(const Ch*, SizeType)` method and call it on the instances returned by the `IRemoteSchemaDocumentProvider` implementation, but it seems inelegant to introduce a setter method to a class that is currently effectively frozen after construction.

## Compatible API changes

* `GenericSchemaDocument`: New method `GetURI()` returns the string that was passed into the new two arguments of the constructor.
* `GenericSchemaValidator`: three new helper typedefs `SValue`, `StringRefType` and `ValueType`.
* `GenericSchemaValidator`: New methods `GetError()` return the error object produced by the most recent validation.
* `GenericSchemaValidator`: A bunch of macro-generated methods returning constant strings for keywords used in error objects.
* `SchemaValidatingReader`: New typedef `ValueType`.
* `SchemaValidatingReader`: New method `GetError()`.

## Internal API changes

* `internal::Schema`: New methods `GetURI()` and `GetPointer()` return the URI of the schema document and the pointer to the schema, respectively.
* `internal::IValidationErrorHandler`: New templated internal interface.
* `internal::Schema`: New public typedef `ErrorHandler`.
* `internal::Schema`: Many validation helper methods now take an additional `ErrorHandler&` argument. Before returning `false` to signal a validation error, they now also call an appropriate method through this interface.
* `GenericSchemaValidator` now implements `internal::IValidationErrorHandler`.
* `GenericSchemaValidator`: Private constructor for parallel validators now receives the base path into the document being validated, to be able to produce accurate error locations.

## Behavior changes

* In case some part of a document violates a remote schema, `GetInvalidSchemaPointer` now returns the pointer relative to that schema, rather than an empty pointer (#1064).
  * This is not absolutely necessary for the feature. If you are opposed, I will revert to the previous behavior.


yurikhan:
I will look into the CI failure.

yurikhan:
Okay, the CI found some C++03 compilation errors, an integer conversion warning, and a real bug concerning a dangling pointer after `SchemaDocument` move construction. I have fixed them all.

@miloyip: Please review.

miloyip:
Just wondering, is it possible to put `ErrorHandler&` in `SchemaValidationContext`, which may remove argument passing overheads?
Besides, I may still need more time to review this big PR.

miloyip:
BTW, I would suggest inserting the markdown documentation to existing schema.md.

yurikhan:
> Just wondering, is it possible to put ErrorHandler& in SchemaValidationContext, which may remove argument passing overheads?

It probably is. Context already keeps a pointer to the Factory, and ErrorHandler is just another interface of the same object. I will try that this weekend.

> Besides, I may still need more time to review this big PR.

By all means, take your time.

> I would suggest inserting the markdown documentation to existing schema.md.

Okay.

----

After a few days of experimental use, I am also inclined to flatten all the `allOf` error groups. It only makes the error object structure more complex, without giving any benefits to the reader. (For `anyOf` and `oneOf`, on the other hand, it is important to know, for each branch separately, how it was violated.)

The motivating use case is this:

```
{ "definitions": {
    "Base": {…base schema keywords…}
  },
  "allOf": [{"$ref": "#/definitions/Base"}],
  …derived schema keywords…
}
```

If the derived schema is violated, we get a simple error object:

```
{…derived schema violations…}
```

If the base schema is violated, we get this:

```
{"allOf": {"errors": [{…base schema violations…}], …}}
```

Especially for longer inheritance chains, that becomes annoying. I think it will be better to simplify the latter into:

```
{…base schema violations…}
```

yurikhan:
I do not know if keeping the ErrorHandler reference in Context is better for performance. But it’s surely neater.

miloyip:
I suggest adding an addition `SchemaDocument()` constructor to include the URI, in order to make it backward compatible.

yurikhan:
I can do that, but then the users of `SchemaDocument` who pass an `IRemoteSchemaDocumentProvider` will not be forced to pass the schema URI. As a result, they will get inaccurate schema references.

In the case where `remoteProvider` is not used, the new interface is backward-compatible.

Should I add the new constructor despite it being easy to use incorrectly?

miloyip:
Thanks

PR #1079, opened 2017-10-03
Make RapidJSON_INCLUDE_DIR non-blank in Config.cmake

captaincrutches:
If I want to use RapidJSON as a submodule for my own project, the currently output RapidJSONConfig.cmake has a blank string for `RapidJSON_INCLUDE_DIR`, making it necessary to include its headers indirectly (in my case, with `${RapidJSON_SOURCE_DIR}/include`).  This change simply sets a value for those variables, so I can use my submodule more easily.


miloyip:
I am not very familiar with CMake actually. Will this affect other current users?

captaincrutches:
Unless current users rely on `RapidJSON_INCLUDE_DIR` being blank in the build tree, I don't see why it would... but if someone knows of a use case like that, I'd be interested to hear it.  This shouldn't affect the install tree if I'm reading the file correctly, so only users who are actually building from source and looking for the package in the source directory (like in a submodule) should see this change.  A `make install` just copies the `RapidJSONConfig.cmake` from the install tree, which doesn't get hit by this.

miloyip:
OK. Thanks

PR #1082, opened 2017-10-05
Use schema allocator for regex

svart-riddare:
Class _internal::GenericRegex<EncodingType, AllocatorType>_ is used by _internal::Schema<SchemaDocumentType>_.

The template parameter for _AllocatorType_ is left to it's default value. This pull request uses Schema::AllocatorType as template parameter and initializes the regex objects created with the schema allocator.

The goal is to use the same (custom) allocator than the one used for _SchemaDocument_. Without this patch, the allocator used is a new instance of the default  _CrtAllocator_, regardless of the allocator used for the _SchemaDocument_.

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -935,7 +935,7 @@ private:
     };
 
 #if RAPIDJSON_SCHEMA_USE_INTERNALREGEX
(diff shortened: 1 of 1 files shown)


miloyip:
Thanks

PR #1083, opened 2017-10-06
Removing Klocwork issues from schema.h

piotr-kaminski-intel:
Removing Klocwork static code analysis critical issues:

line 358:
'this->notValidatorIndex_' might not be initialized in this constructor.
line :412
Pointer 'schemaDocument' checked for NULL at line 412 may be passed to function and may be dereferenced there by passing argument this to function 'CreateSchema' at line 419. Also there are 7 similar errors on lines 467 479 511 523 533 538 549.


miloyip:
Thanks

Issue #1104, opened 2017-10-27
Compilation on Solaris around fread() and fwrite()


clach04:
Different to issue #686.

Error:

    "/...../rapidjson/filereadstream.h", line 71: Error: The function "fread" must have a prototype.
    1 Error(s) detected.

With testcase:

worked with [~dupbr01]. Under Solaris fread() is included from iso/stdio_iso.h which declares fread() in the {{std}} namespace (as one would expect).

Updating rapidjson to use std::fread works (not tried fwrite() yet).

Small test case, test.cpp

{code}
#include "rapidjson.h"
#include "filereadstream.h"
#include "filewritestream.h"
{code}

    CC -I /tmp/cmc/rapidjson/ test.cpp

debug:
    CC -I /tmp/cmc/rapidjson/ test.cpp -E > preprocessed_test.cpp

clach04:
I have a small patch for this.

miloyip:
Please make a PR. Thank you.

miloyip:
Oh. Just saw that.

PR #1105, opened 2017-10-27
Fix issue #1104 Solaris compilation errors fread()/fwrite()

clach04:
Explicit std name space for fread() and fwrite().


miloyip:
Thank you.

Issue #1108, opened 2017-11-02
GenericPointer::Append() and custom allocator


svart-riddare:
Greetings,

I am using `GenericSchemaDocument<GenericValue<UTF8<>, MyOwnAllocator>, MyOwnAllocator>` and it turns out tham some `GenericPointer` used by the schema creates new instance of the allocator instead of the one I provided to the constructor of `GenericSchemaDocument`.

The reason lies in method `GenericPointer GenericPointer::Append(const Token& token, Allocator* allocator = 0) const`, where the last line is `return r`. This return triggers the following calls (at least given my current compilation options, some compilers should be able to optimize out the copy I guess):
- `GenericPointer::GenericPointer(const GenericPointer& rhs, Allocator* allocator = 0)`
- `GenericPointer& GenericPointer::operator=(const GenericPointer& rhs)`
- `Ch* GenericPointer::CopyFromRaw(const GenericPointer& rhs, size_t extraToken = 0, size_t extraNameBufferSize = 0)`

Since the copy constructor is called implicitely, the default valued `allocator` parameter is left to zero and a thus a new instance of the allocator is created in `CopyFromRaw` function.

I could have proposed a pull request for this issue, but there are many ways to resolve this problem, and I am not sure which one is the best.

1. In the copy constructor, replace initialization of `allocator_(allocator)` by `allocator_(allocator ? allocator : rhs.allocator_)`
2. In the assigment operator, add line `if (!allocator_) allocator_ = rhs.allocator_;`
3. In the `CopyFromRaw` method, replace `if (!allocator_) ownAllocator_ = allocator_ = RAPIDJSON_NEW(Allocator());` by `if (!allocator_) allocator_ = rhs.allocator_`, assuming the rhs allocator has been set.

Regards,
Étienne Dupuis

PR #1122, opened 2017-11-23
Suggestion for issue #1108

svart-riddare:
Solves #1108. The default copy constructor of GenericPointer will use the allocator of the copied object. The extra copy constructor that takes an allocator as a parameter is distinct if someone really wants to create a copy with a null allocator.

See  #1108 for details and other possible solutions.


miloyip:
Can you also add an unit test for enforcing this proposed behavior?

svart-riddare:
Unit tests added, for copy constructor and assignment operator. As you will seen, the two tests added emphasize the fact that the copy constructor reuses the allocator of the copied object but not the assignment operator (that was option 1 in my comment on issue #1108).

Of course, we can also implement option 2, in which the copy constructor and the assignment operator have the same behavior regarding the allocator. 

Whichever you prefer.

PR #1124, opened 2017-11-24
CMake cleanups

DerDakon:
- minor cleanups
 - fix crosscompiling by not using -march=native
 - let CMake handle C++11 compiler flag if possible


miloyip:
Thanks

PR #1135, opened 2017-12-12
Added max length parameter for GenericStringStream

jonatanolofsson:
To protect the reader from reading too far in a non null-terminated buffer.

### DIFF
--- a/include/rapidjson/stream.h
+++ b/include/rapidjson/stream.h
@@ -18,6 +18,7 @@
 #define RAPIDJSON_STREAM_H_
 
 #include "encodings.h"
+#include <string>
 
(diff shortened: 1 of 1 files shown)


jonatanolofsson:
I have not yet had the opportunity to verify this commit yet, but I'm putting it here as an implementation example for review and discussion.

pah:
This is a duplicate of #789.

You might want to check out #158 or directly use the provided [`MemoryStream`](http://rapidjson.org/structrapidjson_1_1_memory_stream.html).

jonatanolofsson:
Sorry I missed that, I did try to search =)

Thanks!

PR #1137, opened 2017-12-12
Avoid inheritance from std::iterator

### LINKED ISSUE #1131: std::iterator deprecation
'std::iterator' has been deprecated in C++17 and was never required in the first place. MSVC as of 14.12 compiler version is now warning of the deprecation:

> Warning	C4996
> 
> 'std::iterator<std::random_access_iterator_tag,const rapidjson::GenericMember<Encoding,Allocator>,ptrdiff_t,_Ty *,_Ty &>::difference_type': warning STL4015: The std::iterator class template (used as a base class to provide typedefs) is deprecated in C++17. (The <iterator> header is NOT deprecated.) The C++ Standard has never required user-defined iterators to derive from std::iterator. To fix this warning, stop deriving from std::iterator and start providing publicly accessible typedefs named iterator_category, value_type, difference_type, pointer, and reference. Note that value_type is required to be non-const, even for constant iterators.
> 
> You can define _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.	

All the warnings are coming from 'document.h' file, I hope we can get a fix soon enough. As it was never necessary to use std::iterator, there is no breaking changes to update it in conformance to the standard.

pah:
Instead of inheriting from the deprecated `std::iterator` template, define the member typedefs needed for [`std::iterator_traits`](http://en.cppreference.com/w/cpp/iterator/iterator_traits) directly.

Closes #1131.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -45,7 +45,7 @@ RAPIDJSON_DIAG_OFF(terminate) // ignore throwing RAPIDJSON_ASSERT in RAPIDJSON_N
 #endif // __GNUC__
 
 #ifndef RAPIDJSON_NOMEMBERITERATORCLASS
-#include <iterator> // std::iterator, std::random_access_iterator_tag
+#include <iterator> // std::random_access_iterator_tag
 #endif
 
 #if RAPIDJSON_HAS_CXX11_RVALUE_REFS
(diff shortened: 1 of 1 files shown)


miloyip:
Thank you very much

Issue #852, opened 2017-02-14
RawNumber() inconsistent about quotes


bbeuning:
When reading, RawNumber() does not look for quotes.
But when writing, RawNumber() outputs quotes (because it calls WriteString() internally).
We changed our code to use RawValue() so output does not have quotes.

We use RawNumber() for monetary values where round-off is completely unacceptable.

Thanks,

BGC-nglass:
Also came across this when sending Reader events back into Writer, using kParseNumbersAsStringsFlag to preserve precision and was surprised to find numbers were being quoted.  

Can override as follows in derived writer implementation, which is basically what bbeuning says above.
```
    bool RawNumber(const Ch* str, rapidjson::SizeType length, bool copy = false) {
        (void)copy;
        Prefix(rapidjson::kNumberType);
        return EndValue(WriteRawValue(str, length));
    }
```
This is fine for Reader->Writer events, but wonder if you would want input checking to validate numeric digits if it were to be called in isolation.

Issue #704, opened 2016-08-02
PrettyWriter::RawNumber


deetmo:
Raw number is written as string - inside quotation marks.

Problematic fragment:
`return Base::WriteString(str, length);`

Solution:
`return Base::WriteRawValue(str, length);`

kolomenkin:
All Writers in RapidJSON have a bug of this kind: PrettyWriter, Writer
Their method RawNumber is broken.

The problem was first reported in 2016 (10 years ago). The problem is very simple to understand and fix.
Some PRs were proposed in 2016, but were closed without merging.

Latest RapidJSON code is still broken in 2026. Alas!

PR #1155, opened 2018-01-08
Write RawNumber without quotes

h46incon:
This PR try to fix issue that `Writer.RawNumber()` writes number as string with quotes (#560 #704 #852). Some Considerations:
1. Raw number should contain characters in range [1,127], so it's safe that cast it to `OutputStream::Ch`.
2. No futher checking as `WriteRawValue()`.

### DIFF
--- a/include/rapidjson/writer.h
+++ b/include/rapidjson/writer.h
@@ -198,7 +198,7 @@ public:
         RAPIDJSON_ASSERT(str != 0);
         (void)copy;
         Prefix(kNumberType);
-        return EndValue(WriteString(str, length));
+        return EndValue(WriteRawNumber(str, length));
     }
 
     bool String(const Ch* str, SizeType length, bool copy = false) {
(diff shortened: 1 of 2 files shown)


h46incon:
The CI failed because throw in non-throwing function, e.g. 
```cpp
explicit GenericValue(Type type) RAPIDJSON_NOEXCEPT
{
       RAPIDJSON_ASSERT(type >= kNullType && type <= kNumberType);
      //...
```

h46incon:
@miloyip The encoding is more complex than I thought, like #1179.

chrisvroberts:
Hi @miloyip, @h46incon.

I am trying to understand why this has been closed. looking at master HEAD RawNumber in writer.h still calls WriteString meaning quotes will be added to numbers. This breaks the whole idea of the SAX reader and the writer having the same interface because the JSON output by RawNumber in the writer won't match the input read in by the reader and presented in the RawNumber call back.
```
1234 -> SAX reader -> RawNumber -> "1234"
"1234" -> RawNumber -> writer ->"\"1234\""
```
Can RawNumber(p, l, b) not just call RawValue(p, l, kNumberType)?

pitrou:
Indeed the behaviour of `RawNumber` is unexpected and is also biting us.
If the concern is to avoid breaking compatibility, perhaps add a boolean flag for the desired behaviour?

mapleFU:
I meet the same problem, seems this project might lacking activity, I think using another lib or extend from current writer is ok...

PR #1161, opened 2018-01-11
Bug when switching to std regex

xiaoPierre:
I could not switch to std regex after defining the two variables as in documents. Then I try to fix it in schema.h.


miloyip:
Thanks

PR #1172, opened 2018-01-29
Fix #1170. Structs are initialized for archivertest.

### LINKED ISSUE #1170: Compile error because of -Werror=effc++ is on.
Some of the errors log are pasted below 

`/home/hui/rvbust/source/rapidjson/example/archiver/archivertest.cpp:8:8: error: ‘Student::name’ should be initialized in the member initialization list [-Werror=effc++]
 struct Student {
        ^
/home/hui/rvbust/source/rapidjson/example/archiver/archivertest.cpp:8:8: error: ‘Student::age’ should be initialized in the member initialization list [-Werror=effc++]
/home/hui/rvbust/source/rapidjson/example/archiver/archivertest.cpp:8:8: error: ‘Student::height’ should be initialized in the member initialization list [-Werror=effc++]
`
I am on Ubuntu 16.04, gcc version is 5.4.0. Turn off effc++ option will fix the issue. But wonder why this is happening.

justinjereza:
struct Student is not a POD since it has an std::string member. Just initializing each member will cause an error since only a default contructor with no parameters and a copy constructor will be implicitly defined thus preventing brace initialization. Defining constructors for Student seems to be the best thing to do.

### DIFF
--- a/example/archiver/archivertest.cpp
+++ b/example/archiver/archivertest.cpp
@@ -35,7 +35,7 @@ void test1() {
 
     // Serialize
     {
-        Student s = { "Lua", 9, 150.5, true };
+        Student s("Lua", 9, 150.5, true);
 
         JsonWriter writer;
         writer & s;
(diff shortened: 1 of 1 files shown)


miloyip:
Fixed in https://github.com/Tencent/rapidjson/commit/82b5c425676d2b07e88c1a8d243ea8284cc21a1d
Thank you.

Issue #1086, opened 2017-10-11
-Wclass-memaccess triggered by GCC7


h-2:
I am using RapidJSON via Cereal and it triggers the above warning, please see USCiLab/cereal#434

Can you fix this warning or locally suppress it?

Thank you very much!

miloyip:
I think suppression is the way to go. I do not have that build environment. Can you make a PR?

marehr:
I think this was fixed and can be closed?

marehr:
Hmm, #1205 indicates that this is not fixed yet.

PR #1182, opened 2018-02-19
Ignore GCC 8 warnings.

Romain-Geissler-1A:
Hi,

This should fix issue #1086 that you advised we fix by ignoring the warning rather than by changing the code logic.

Cheers,
Romain


miloyip:
Thank you

PR #1186, opened 2018-03-02
add find_package(Rapidjson) module for cmake

owent:
增加一个用于cmake查找rapidjson的模块文件。我们好几个项目都有用到。
不知道 @miloyip 大大有没兴趣收录

### DIFF
--- a/contrib/cmake/FindRapidjson.cmake
+++ b/contrib/cmake/FindRapidjson.cmake
@@ -1,0 +1,71 @@
+﻿#.rst:
+# FindRapidjson
+# --------
+#
+# Find the native rapidjson includes and library.
+#
+# IMPORTED Targets
+# ^^^^^^^^^^^^^^^^
+#
+#
+# Result Variables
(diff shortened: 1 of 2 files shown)


steveire:
This is the wrong thing to do. Instead, a RapidJsonConfig.cmake flie should be provided along with a RapidJsonConfigVersion.cmake. See https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html

ZhenshengLee (approved):
Great Job.

steveire:
Please see my cmake link above. This is the wrong thing to do. You should provide a `Config.cmake` file instead.

ZhenshengLee:
> Please see my cmake link above. This is the wrong thing to do. You should provide a `Config.cmake` file instead.
> 
> What you are doing here is "putting the treasure map beside the treasure". See https://youtu.be/GJ0kMsLbk6Q?t=1249

Thanks you for your reply, I will have a check.

PR #1191, opened 2018-03-06
Support long and unsined long as int and unsigned on Microsft platforms

Sumoren:
We are still using long and unsigned long as s32 and u32 on Microsoft platforms and it was not supported by the TypeHelper.

I added support for long and unsigned long to be Int and UInt when _MSC_VER is definded.
I also added static assert to validate long and unsigned long size in case one day Microsoft change its mind.


Sumoren:
Sorry for the two commits for the unit test. I saw my error only after I pushed to my github repo. I suppose you can squash the commits when you accept a pull request? I'm more used to gitlab UI,

miloyip:
Thanks

Issue #1184, opened 2018-02-23
IsFloat and IsDouble not consistant with GetFloat and GetDouble


Sumoren:
If in the json a number value is a int like:
{
    "Number": 123
}
IsFloat(), Is<float>(), IsDouble() and Is<double>() return false while 123 is a valid float or double value.
Even GetFloat(), Get<float>(), GetDouble() and Get<double>() return the value correctly.

It limit what can be done in template since we cannot check Is<T>() to know if Get<T>() will return a valid value or assert without doing specialization for float and double to check IsNumber instead.

I would expect IsXXX to return true if it matching GetXXX can return a valid value.

miloyip:
`GetDouble()` will be valid when `IsNumber() == true`.

Currently the API `IsDouble()` is for querying whether the value is stored in `double`. It does not handle implicit conversion in `GetDouble()`.

I think it may need template specialization for your need, which may specialized for a templated function, e.g. `IsConvertibleTo<double>(const Value& v) { return v.IsNumber(); }`.

Sumoren:
From your answer I cannot even tell if you think the current behavior is correct or not.
Yes I can do a templated funciton with specialization for double/float and that what I did. But I would expect the API to better handle double.

About this comment:
>GetDouble() will be valid when IsNumber() == true.

Did you means in a future path? Because that not true. If you write this program:

    int main()
    {
        const char* json =
    R"({
        "Number": 123
    })";

        rapidjson::Document doc;
        doc.Parse(json);

        rapidjson::Value& number{ doc.FindMember("Number")->value };

        std::cout << std::boolalpha;
        std::cout << "IsNumber: " << number.IsNumber() << std::endl;
        std::cout << "Is<double>: " << number.Is<double>() << std::endl;
        std::cout << "IsDouble:" << number.IsDouble() << std::endl;

        return 0;
    }

You get this output:

    IsNumber: true
    Is<double>: false
    IsDouble:false

Also about your comment:
>It does not handle implicit conversion in GetDouble().

It look to me like GetDouble do implicit conversion from int, uint, int64 and uint 64:

    double GetDouble() const {
        RAPIDJSON_ASSERT(IsNumber());
        if ((data_.f.flags & kDoubleFlag) != 0)                return data_.n.d;   // exact type, no conversion.
        if ((data_.f.flags & kIntFlag) != 0)                   return data_.n.i.i; // int -> double
[...]

PR #1192, opened 2018-03-06
TypeHelper::Is for double and float return true when value is a number

Sumoren:
Really not sure if you agree with this one after the discussion on https://github.com/Tencent/rapidjson/issues/1184. But I send it anyway.

This change allow rapidjson::GenericValue::Is to return true when rapidjson::GenericValue::Get is able to return a value without wrapping the call to Is in template to be able to write specialization for float and double.

It also keep the current behavior for IsDouble and IsFloat to return false for value like:
`"Number": 1234`


miloyip:
It slightly changes the behavior of the helper but it should be fine.

miloyip:
I wait for some time to see if anyone comment on this.

Sumoren:
Hi,
Just to know if you have decided if you want this change? If not I will not maintains it on top of the master.
Thank you

PR #1199, opened 2018-03-12
Fixes #1198

### LINKED ISSUE #1198: Test failure: SchemaValidator.TestSuite with additionalItems.json
Win10, VS2010 x64, rapidjson master.

unit tests fail at SchemaValidator.TestSuite with additionalItems.json, at

Ch* CopyFromRaw(const GenericPointer& rhs, size_t extraToken = 0, size_t extraNameBufferSize = 0)

around lines

size_t nameBufferSize = rhs.tokenCount_; // null terminators for tokens
        for (Token *t = rhs.tokens_; t != rhs.tokens_ + rhs.tokenCount_; ++t)
            nameBufferSize += t->length;

local variables:
image

rhs.tokens_ seems to be uninitialized, hence subsequent Malloc call fails causing SEH exception.

kachanovskiy:
Problem: unit test fails unit at SchemaValidator.TestSuite with additionalItems.json on Windows build with VS2010 (possibly with others)

Resolution: changed return value format.

This PR is supposed to repair Appveyor Windows builds that seem to be half-broken since some time.

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -1860,7 +1860,12 @@ public:
 
     //! Gets the JSON pointer pointed to the invalid value.
     PointerType GetInvalidDocumentPointer() const {
-        return documentStack_.Empty() ? PointerType() : PointerType(documentStack_.template Bottom<Ch>(), documentStack_.GetSize() / sizeof(Ch));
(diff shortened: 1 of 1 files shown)


miloyip:
Thank you very much.

PR #1202, opened 2018-03-14
Change long/ulong as int/uint on MSC unit tests to be more inline wit…

Sumoren:
I saw in pull request https://github.com/Tencent/rapidjson/pull/1191 that my tests didn't cover all calls to type helper. So I updated the tests to me more inline with the other unit test for templated function just above the tests I added.


miloyip:
Thank you for improving the test coverage.

PR #1207, opened 2018-03-19
Ensure a preprocessor directive is only applied to MSVC.

vim2meta:
The `#pragma intrinsic` preprocessor directive is only implemented on the MSVC compiler. However, the Intel compiler defines `_MSC_VER` to be compatible with MSVC. This results in the following warning when compiling using Intel: `unimplemented pragma ignored` every time the `biginteger.h` file is included.


miloyip:
Thanks

PR #1210, opened 2018-03-23
add remote ref pointer to schemaMap_

foxtacles:
Hello,

I think I've encountered a bug. I noticed wrong results when validating a schema referencing a remote schema, which references another remote schema. Here is an example: https://gist.github.com/foxtacles/0deaa03be09cf9820e49cc590ee020c6

`x.json` references `y.json` references `z.json`. The validation should fail (`UK` is not part of the `enum`). However, it succeeds in the current `master` version of RapidJSON.

I found this is because in the `HandleRefSchema` function, `remoteDocument->GetSchema(pointer)` returns a null pointer. In the case of `y.json`, the pointer `/definitions/country` is not part of the internal `schemaMap_`.  Since RapidJSON silently ignores this, the validation succeeds.

I'm not sure if this is intended behaviour - maybe I'm using the library in some wrong way, but I've attached a fix that resolves this problem (storing the schema pointer with a non-owning pointer to the remote schema).

Thanks for your great work!

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -1673,6 +1673,7 @@ private:
                                 if (const SchemaType* sc = remoteDocument->GetSchema(pointer)) {
                                     if (schema)
                                         *schema = sc;
+                                    new (schemaMap_.template Push<SchemaEntry>()) SchemaEntry(source, const_cast<SchemaType*>(sc), false, allocator_);
                                     return true;
                                 }
                             }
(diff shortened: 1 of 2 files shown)


miloyip:
I think you are right.

Can you also put your test code into https://github.com/Tencent/rapidjson/blob/master/test/unittest/schematest.cpp ? (Need using `std::map` instead of `std::unordered_map` for C++03)

foxtacles:
@miloyip I added a unit test (also verified that it fails in master). I didn't use `std::map` since that would introduce a new dependency, so I did something simple instead, but it's very similar.

miloyip:
Thank you very much

PR #1213, opened 2018-04-03
add string_view support

antkudr:


miloyip:
I am not familiar with this C++ feature. Can @pah or others help to review on this?

Bu11etmagnet:
`std::basic_string_view<>` has the same purpose as `rapidjson::GenericStringView<>` : a reference to a string owned by someone else.

antkudr:
or `boost::string_view`, `QStringView`

pah:
In general, support for `string_view` is a nice addition.

The current PR (silently) does not allow defining both `RAPIDJSON_HAS_STDSTRINGVIEW` and `RAPIDJSON_HAS_STDSTRING` (e.g. the unittests will fail in this case).

I think, it would be better to really make these features mutually exclusive and only consider `RAPIDJSON_HAS_STDSTRING`, if `RAPIDJSON_HAS_STDSTRINGVIEW` is **not** defined. And in turn, implying `RAPIDJSON_HAS_STDSTRING`, if `RAPIDJSON_HAS_STDSTRINGVIEW` **is** defined.

This can be achieved by adding something like the following to `GenericValue`, et.al:

```cpp
#if RAPIDJSON_HAS_STDSTRINGVIEW
  typedef std::basic_string_view<Ch>   StdStringLikeArg;
#elif RAPIDJSON_HAS_STDSTRING
  typedef const std::basic_string<Ch>& StdStringLikeArg;
#endif 
```
With this, we don't need to duplicate the APIs (and their implementations)  in `GenericValue`/`GenericObject`/`GenericDocument`, but can just provide the overload taking a `StdStringLikeArg`, e.g.
```cpp
#if RAPIDJSON_HAS_STDSTRING || RAPIDJSON_HAS_STDSTRINGVIEW
     bool HasMember(StdStringLikeArg name) const {return FindMember(name) != MemberEnd(); }
#endif
```
The `TypeHelper` specializations are still needed for both `std::basic_string`and `std::basic_string_view`.

Bu11etmagnet:
I don't get this. Why can't there be independent overloads for both `std::string` and `std::string_view`?

Besides, `RAPIDJSON_HAS_STDSTRINGVIEW` should imply `RAPIDJSON_HAS_STDSTRING`.

csw:
I think there really should be overloads for both `std::string` and `std::string_view`; they have different purposes and I'd use RapidJSON with both.

I've rebased @antkudr's branch onto master and made a few changes so that the `std::string_view` support coexists with `std::string` support and passes all the tests; see csw/rapidjson@ddd8e6c8ed3c68a33c53afe6d232352e5a49b1ee. Should I open a separate PR for that branch?

pah:
> I think there really should be overloads for both std::string and std::string_view; they have different purposes and I'd use RapidJSON with both.

What purpose does the `std::string` overload have, if you can call the same via `std::string_view`? Can you elaborate?

csw:
That’s a good point, the implicit conversion from `std::string` to `std::string_view` should mean that support for the latter would cover the `std::string` case by itself. I hadn’t thought about that aspect. I’d want to see both cases tested explicitly, but that approach makes perfect sense to me. I’m happy to rework my changes to do it that way, if that’d be helpful.

antkudr:
You will break source compatibility.

class A
{
public:
    A() = default;
    operator std::string() const { return "string"; }
};

void f(std::string_view s)
{
    std::cout << s;
}

/*void f(const std::string& s)
{
    std::cout << s;
}
*/

void main()
{
A a;
f(a);
}

pah:
> You will break source compatibility.

My proposal above would still allow using`std::string` instead of `std::string_view`.  Projects with code like the above can stick to `RAPIDJSON_HAS_STDSTRING`. 

I would feel, that the above (somewhat contrived) example would not warrant the explosion of overloads.  Real use cases for custom string classes are likely adjusted to support `std::string_view` conversion instead of a conversion to `std::string` anyway, when moving to C++17.

PR #1217, opened 2018-04-06
Fix compilation on windows with clang

DirtYiCE:
Currently rapidjson produces many warning messages when compiling it with clang-cl on windows using MSVC 2017 libraries. The problem is that in this case `_MSC_VER` and `__clang__` are both defined, and the headers doesn't handle this case correctly.

The file `test/unittest/schematest.cpp` had a compile error under gcc with `-Werror=effc++` since `include/rapidjson/internal/regex.h` didn't pop the diagnostic in case of gcc.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -26,15 +26,13 @@
 #include <limits>
 
 RAPIDJSON_DIAG_PUSH
-#ifdef _MSC_VER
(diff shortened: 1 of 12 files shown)


steveire:
Please also integrate the change in https://github.com/Tencent/rapidjson/pull/1231 into this one.

PR #1222, opened 2018-04-17
Allow user configuration of default chunk capacity.

k3rb3ros:
Added macro RAPIDJSON_ALLOCATOR_DEFAULT_CHUNK_CAPACITY to allow default chunk capacity to be lowered for embedded devices with small memory.


miloyip:
OK thanks

Issue #1015, opened 2017-07-11
`_LIBCPP_VERSION` is not reliable for clang on windows


xylosper:
RapidJSON checks `_LIBCPP_VERSION` to determine C++11 support such as rvalue reference.

However, on Windows, clang supports most of all C++11/C++14 features but it does not define `_LIBCPP_VERSION` because it does not use libc++.

Athosvk:
A late second on this issue. It's quite easy to hack around this, defining RAPIDJSON_HAS_CXX11_RVALUE_REFS ourselves, but it's not so nice of a solution. Clang honestly shouldn't be related to _which_ standard library is used. I doubt it would hurt making it agnostic to the standard library vendor in the first place (as long as you check the version).

miloyip:
Any suggestion?

Bu11etmagnet:
`__clang_major__` , `__clang_minor__`, http://en.cppreference.com/w/cpp/experimental/feature_test

Athosvk:
I'm not certain if any other standard lib vendors have defines for their standard lib version, so I guess that's a tough one. That being said, is it really necessary to double check for the standard library version if you're already guaranteed the C++ version is correct? I mean if you have the right C++version but not the correct standard lib version, you're pretty far off a correct setup in the first place. I would, honestly, remove the check for the lib version in case there's no way to check this for other standard lib implementations, simply due to the behaviour now being 'incorrect'.

Bu11etmagnet:
Note that the standard C++ library is not necessarily tied to the compiler (e.g. clang - any version - on CentOS 7 uses the system C++ library, which is the libstdc++ which belongs to the default compiler - GCC 4.8).

PR #1231, opened 2018-04-24
Use rvalue refs with clang-cl

steveire:
Fix natural usage with

 https://clang.llvm.org/docs/UsersManual.html#clang-cl


steveire:
Note that unlike https://github.com/Tencent/rapidjson/pull/1217 which resolves warnings, this PR fixes errors encountered in simple user code such as 

    Value x(1234);
    x = Value(4321);

steveire:
This is mentioned in https://github.com/Tencent/rapidjson/issues/1015.

PR #1241, opened 2018-05-04
compatiblities with Itanium Hp-Ux platform

mariusz-jasinski:
I have changed some files after compiling on Itanium Hp-Ux platform.
I have tested suscessfull this changing on HP-UX Itanium B.11.31 U 9000/800 64-bits, compiler: HP C/aC++ B3910B A.06.28.
Could You merge this compatiblities on master ?

### DIFF
--- a/include/rapidjson/rapidjson.h
+++ b/include/rapidjson/rapidjson.h
@@ -255,7 +255,7 @@
 
 //! Whether using 64-bit architecture
 #ifndef RAPIDJSON_64BIT
-#if defined(__LP64__) || (defined(__x86_64__) && defined(__ILP32__)) || defined(_WIN64) || defined(__EMSCRIPTEN__)
+#if defined(__LP64__) || (defined(__x86_64__) && defined(__ILP32__)) || defined(_WIN64) || defined(__EMSCRIPTEN__) || (defined(__hpux) && defined(__ia64))
 #define RAPIDJSON_64BIT 1
 #else
 #define RAPIDJSON_64BIT 0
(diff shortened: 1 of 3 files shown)


include/rapidjson/writer.h
```diff
@@ -46,6 +46,11 @@ RAPIDJSON_DIAG_PUSH
 RAPIDJSON_DIAG_OFF(4127) // conditional expression is constant
 #endif
 
+#ifdef __hpux
+// disable warning unreachable-code
+#pragma diag_suppress 2111
+#endif
```
pah: You should use `diag_push`, `diag_pop` to restore the original selection of the application.

Does the HP-UX compiler support something like `_pragma`/`_Pragma`?  In this case, you can add support for the `RAPIDJSON_DIAG_*` macros on this platform as well, see `rapidjson.h`.

include/rapidjson/rapidjson.h
```diff
@@ -255,7 +255,7 @@
 
 //! Whether using 64-bit architecture
 #ifndef RAPIDJSON_64BIT
-#if defined(__LP64__) || (defined(__x86_64__) && defined(__ILP32__)) || defined(_WIN64) || defined(__EMSCRIPTEN__)
+#if defined(__LP64__) || (defined(__x86_64__) && defined(__ILP32__)) || defined(_WIN64) || defined(__EMSCRIPTEN__) || (defined(__hpux) && defined(__ia64))
```
pah: Isn't `__ia64` enough?

Issue #1253, opened 2018-05-11
Parsing "2e308" in full precision generates NAN value and kParseErrorNone


StilesCrisis:
2e308 is slightly beyond the valid range of a double.
This is a slight variation on Issue #849

StilesCrisis:
Unit test demonstrating the behavior is now in #1254

PR #1254, opened 2018-05-11
Unit Test: Parsing "2e308" in full precision generates NAN value and kParseErrorNone

StilesCrisis:


StilesCrisis:
Adding a unit test for #1253

Issue #1255, opened 2018-05-12
Parsing "6223372036854775296.1701512723685473547372536......" causes read from uninitialized memory


StilesCrisis:
Parsing the input string "6223372036854775296.170151272368547354737253685475529337203685468547752975223373720170151233720097201372368547312337203687203685423685473724737236837236775297525854775297525472975254737236873720170151235473783723677529752585477247372368372368547354737253685475529752Ë18 9Œ" causes the RapidJSON parser to read from uninitialized memory in rapidjson::internal::BigInteger::Compare.

StilesCrisis:
Unit test demonstrating the behavior is now in #1256

szaszm:
This seems to be fixed by #1290, consider closing the issue

PR #1256, opened 2018-05-12
Unit Test: Parsing "6223372036854775296.1701512723685473547372536......" causes read from uninitialized memory

StilesCrisis:


StilesCrisis:
Adding a unit test for #1255

PR #1261, opened 2018-05-16
Prevent int underflow when parsing exponents

fmalita:
When parsing negative exponents, the current implementation takes
precautions for |exp| to not underflow int.

But that is not sufficient: later on [1], |exp + expFrac| is also
stored to an int - so we must ensure that the sum stays within int
representable values.

Update the exp clamping logic to take expFrac into account.

[1] https://github.com/Tencent/rapidjson/blob/master/include/rapidjson/reader.h#L1690

### DIFF
--- a/include/rapidjson/reader.h
+++ b/include/rapidjson/reader.h
@@ -1632,9 +1632,18 @@ private:
             if (RAPIDJSON_LIKELY(s.Peek() >= '0' && s.Peek() <= '9')) {
                 exp = static_cast<int>(s.Take() - '0');
                 if (expMinus) {
+                    // (exp + expFrac) must not underflow int => we're detecting when -exp gets
+                    // dangerously close to INT_MIN (a pessimistic next digit 9 would push it into
+                    // underflow territory):
+                    //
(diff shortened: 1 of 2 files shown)


include/rapidjson/reader.h
```diff
@@ -1632,9 +1632,18 @@ class GenericReader {
...
                 exp = static_cast<int>(s.Take() - '0');
                 if (expMinus) {
+                    // (exp + expFrac) must not underflow int => we're detecting when -exp gets
+                    // dangerously close to INT_MIN (a pessimistic next digit 9 would push it into
+                    // underflow territory):
+                    //
+                    //        -(exp * 10 + 9) + expFrac >= INT_MIN
+                    //   <=>  exp <= (expFrac - INT_MIN - 9) / 10
+                    RAPIDJSON_ASSERT(expFrac <= 0);
+                    int maxExp = (expFrac + 2147483639) / 10;
```
miloyip, line 1642: See if possible to prevent division.

carnil:
This issue appears to have assigned CVE-2024-38517 .

Issue #784, opened 2016-10-27
Assigning default values defined in JSON Schema


pferdinand:
Any plan of supporting the assignment of default values during the schema validation?

stk-ableton:
I'm interested in this too. Any ideas about the likelihood of adding this?

bogaotory:
This would be really useful. According to [this](https://spacetelescope.github.io/understanding-json-schema/reference/generic.html) description of JSON schema:

> ... though many JSON schema validators simply ignore the default keyword ...

Indeed, so far I have not found a validator (C++) which does this. +1 on adding this feature.

PR #1279, opened 2018-06-01
Avoid `MissingProperty` Error from `SchemaValidator` when a none-zero length value is given as `default` for the property

bogaotory:
In relation to issue #784 : this pull request enables the `SchemaValidator` to recognise the `default` keyword of a JSON schema, so that:

1. When a none-zero string is given as `default` for a property, a `MissingProperty` error is avoided;
2. The actual content of the string given by the `default` keyword is not checked by the schema at all, only the length of the string is concerned;
3. No modification (such as requested by #784) to the actual JSON document object is made. The programmer is still left with the task to check if the property exist in the document, and if and when `default` is needed, what `default` value it has (by reading the schema document object) .

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -441,7 +441,7 @@ public:
         maxLength_(~SizeType(0)),
         exclusiveMinimum_(false),
         exclusiveMaximum_(false),
-        defaultValue_()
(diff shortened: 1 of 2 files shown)


miloyip:
Thank you

Issue #1280, opened 2018-06-03
'operator=' has a non-throwing exception specification but can still throw


mobileben:
I'm using rapidjson for a game running iOS. I'm getting complier warnings related to rapidjson

> 'operator=' has a non-throwing exception specification but can still throw [-Wexceptions]
> 'GenericValue' has a non-throwing exception specification but can still throw [-Wexceptions]

The issue at hand is in code such as this:

    GenericValue& operator=(GenericValue& rhs) RAPIDJSON_NOEXCEPT {
        RAPIDJSON_ASSERT(this != &rhs);
        this->~GenericValue();
        RawAssign(rhs);
        return *this;
    }

The problem is I can't modify ```RAPIDJSON_ASSERT``` to "do nothing" if ```RAPIDJSON_NOEXCEPT``` becomes noexcept since ```RAPIDJSON_ASSERT``` is used in other contexts independent of ```RAPIDJSON_NOEXCEPT```.

It seems perhaps there needs to be a variant (terrible name warning) like ```RAPIDJSON_ASSERT_NOEXCEPT``` that can turn itself off when ```clang``` and ```__has_feature(cxx_noexcept)``` is true

I just pulled the latest and the issue still exists.

Note that this issue also surfaced here:

https://github.com/USCiLab/cereal/issues/439

miloyip:
I think it could be safely changed to

~~~cpp
    if (this != &rhs) {
        this->~GenericValue();
        RawAssign(rhs);
    }
    return *this;
~~~

mobileben:
I'll update and create a pull request. There are still a couple of other instances of this:

    explicit GenericValue(Type type) RAPIDJSON_NOEXCEPT : data_() {
        static const uint16_t defaultFlags[7] = {
            kNullFlag, kFalseFlag, kTrueFlag, kObjectFlag, kArrayFlag, kShortStringFlag,
            kNumberAnyFlag
        };
        RAPIDJSON_ASSERT(type >= kNullType && type <= kNumberType);
        data_.f.flags = defaultFlags[type];

        // Use ShortString to store empty string.
        if (type == kStringType)
            data_.ss.SetLength(0);
    }

and

```
#ifndef RAPIDJSON_DOXYGEN_RUNNING // hide SFINAE from Doxygen
    template <typename T>
    explicit GenericValue(T b, RAPIDJSON_ENABLEIF((internal::IsSame<bool, T>))) RAPIDJSON_NOEXCEPT  // See #472
#else
[...]

PR #1284, opened 2018-06-07
Handle non-throwing exception specifications that can still throw #1280

mobileben:
This is to address https://github.com/Tencent/rapidjson/issues/1280

Note that I have opted a means to control if this is applied or not through RAPIDJSON_ASSERT_THROWS

Example usage in my code setting this is

```
#define RAPIDJSON_ASSERT_THROWS 1

#define RAPIDJSON_ASSERT(x) if (!(x)) throw RapidJsonException();
```

### DIFF
--- a/include/rapidjson/rapidjson.h
+++ b/include/rapidjson/rapidjson.h
@@ -592,6 +592,19 @@ RAPIDJSON_NAMESPACE_END
 //!@endcond
 
 ///////////////////////////////////////////////////////////////////////////////
+// RAPIDJSON_NOEXCEPT_ASSERT
+
+#ifdef RAPIDJSON_ASSERT_THROWS
+#if RAPIDJSON_HAS_CXX11_NOEXCEPT
+#define RAPIDJSON_NOEXCEPT_ASSERT(x)
+#else
+#define RAPIDJSON_NOEXCEPT_ASSERT(x) RAPIDJSON_ASSERT(x)
+#endif // RAPIDJSON_HAS_CXX11_NOEXCEPT
+#else
+#define RAPIDJSON_NOEXCEPT_ASSERT(x) RAPIDJSON_ASSERT(x)
+#endif // RAPIDJSON_ASSERT_THROWS
+
+///////////////////////////////////////////////////////////////////////////////
 // new/delete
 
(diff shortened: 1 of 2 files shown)


include/rapidjson/rapidjson.h
```diff
@@ -591,6 +591,19 @@ RAPIDJSON_NAMESPACE_END
 
 //!@endcond
 
+///////////////////////////////////////////////////////////////////////////////
+// RAPIDJSON_NOEXCEPT_ASSERT
+
+#ifdef RAPIDJSON_ASSERT_THROWS
+#if RAPIDJSON_HAS_CXX11_NOEXCEPT
+#define RAPIDJSON_NOEXCEPT_ASSERT(x)
```
pah: Removing the assertion entirely is not the ideal solution.  I would prefer to 
* allow a separate override of `RAPIDJSON_NOEXCEPT_ASSERT`
* defaulting to 
  * `RAPIDJSON_ASSERT`, if `!RAPIDJSON_HAS_CXX11_NOEXCEPT`
  * plain `assert` (i.e. default of `RAPIDJSON_ASSERT`) otherwise

Something like:
````cpp
//! Assertion (in non-throwing contexts).
 /*! \ingroup RAPIDJSON_CONFIG
    Some functions provide a \c noexcept guarantee, if the compiler supports it.
    In these cases, the \ref RAPIDJSON_ASSERT macro cannot be overridden to
    throw an exception.  This macro adds a separate customization point for
    these cases.

    Defaults to C \c assert() (as \ref RAPIDJSON_ASSERT), if \c noexcept is
    supported, and to \ref RAPIDJSON_ASSERT otherwise.
 */
#ifndef RAPIDJSON_NOEXCEPT_ASSERT
#if RAPIDJSON_HAS_CXX11_NOEXCEPT
#include <cassert>
#define RAPIDJSON_NOEXCEPT_ASSERT(x) assert(x)
#else
#define RAPIDJSON_NOEXCEPT_ASSERT(x) RAPIDJSON_ASSERT(x)
#endif // RAPIDJSON_HAS_CXX11_NOEXCEPT
#endif // RAPIDJSON_NOEXCEPT_ASSERT
````

Issue #1259, opened 2018-05-15
Parsing "8847432036854773723683723..." in full precision mode asserts with "!Sign()" 


StilesCrisis:
Parsing the following number in full-precision mode leads to an assertion in NextPositiveDouble:

8847432036854773723683723677529854735473725368547554755293372036854685477529752529337203685468547770151233720097201372368547312337203687203685423685123372036872036854737247372368372367752975258547752975254729752547372368737201701512354737837236737247372368772473723683723456789012E66

        RAPIDJSON_ASSERT(!Sign());

Issue #1251, opened 2018-05-11
Parsing "128.74836467836484838364836483643636483648e-336" causes a crash


StilesCrisis:
The number "128.74836467836484838364836483643636483648e-336" causes rapidjson::internal::GetCachedPower10 to overflow a signed integer.

Sanitizer report:
third_party/rapidjson/include/rapidjson/internal/diyfp.h:242:47: runtime error: signed integer overflow: 536870911 * 8 cannot be represented in type 'int'

StilesCrisis:
This can lead to a segfault.

Program received signal SIGSEGV, Segmentation fault.
GetCachedPower10 (outExp=<synthetic pointer>, exp=-353) at ~/github/rapidjson/include/rapidjson/internal/diyfp.h:244
244	     return GetCachedPowerByIndex(index);
(gdb) bt
#0  GetCachedPower10 (outExp=<synthetic pointer>, exp=-353) at ~/github/rapidjson/include/rapidjson/internal/diyfp.h:244
#1  StrtodDiyFp (result=<optimized out>, exp=<optimized out>, decimalPosition=<optimized out>, length=<optimized out>, decimals=<optimized out>)
    at ~/github/rapidjson/include/rapidjson/internal/strtod.h:154
#2  StrtodFullPrecision (exp=-336, decimalPosition=3, length=<optimized out>, decimals=0x555555dfb800 "12874836467836484838364836483643636483648", p=<optimized out>, 
    d=<optimized out>) at ~/github/rapidjson/include/rapidjson/internal/strtod.h:259
#3  rapidjson::GenericReader<rapidjson::UTF8<char>, rapidjson::UTF8<char>, rapidjson::CrtAllocator>::ParseNumber<16u, rapidjson::GenericStringStream<rapidjson::UTF8<char> >, ParseDoubleHandler> (this=this@entry=0x7fffffffd920, is=..., handler=...) at ~/github/rapidjson/include/rapidjson/reader.h:1692
#4  0x00005555558653d7 in rapidjson::GenericReader<rapidjson::UTF8<char>, rapidjson::UTF8<char>, rapidjson::CrtAllocator>::ParseValue<16u, rapidjson::GenericStringStream<rapidjson::UTF8<char> >, ParseDoubleHandler> (this=this@entry=0x7fffffffd920, is=..., handler=...)
    at ~/github/rapidjson/include/rapidjson/reader.h:1731
[...]

Issue #1249, opened 2018-05-11
Parsing "0e100" in full precision mode causes UB


StilesCrisis:
Attempting to parse a document containing "0e100" in full precision mode will call DiyFp::Normalize() where f == 0. 

This will cause an infinite loop/soft-lock on platforms that take the plain C implementation, and generate a random value on platforms that use a builtin; __builtin_clzll and _BitScanReverse64 are both implemented with the BSR opcode, which is documented as producing an undefined value.

miloyip:
Sorry I was too busy these few days. Will take a look.
Will u also be possible to suggest fixes on these?

StilesCrisis:
Diyfp is a black box to me. I don't have any suggestions.

miloyip:
Fixed in #1290 and verified in #1250.

PR #1290, opened 2018-06-15
Fix strtod

### LINKED ISSUE #849: Inconsistent reporting of numbers that are too big to be stored in a double

abolz:
Fixes #849, #1249, #1251, #1253, #1256, #1259

The issues were mainly because of unsigned integer overflow and boundary cases. Switched to signed integer arithmetic.

To do:
* This patch currently requires "small inputs" and "small exponents" due to integer overflow. Fixing this probably requires some more changes in `ParseNumber`. (Or simply using int64 instead of int32)
* Some inaccuracies in full-precision mode (due to discarding non-zero bits) should be fixed.


miloyip:
Really thank for your contribution.
There is some valgrind errors such as https://travis-ci.org/Tencent/rapidjson/jobs/392709049#L2555 Can you take a look?

abolz:
Yes, looking into it. But honestly I have no idea whats wrong... Valgrind complains about an uninitialized variable of size 4 in StrtodBigInteger.

abolz:
BTW could you cancel all the running appveyor builds?

abolz:
The valgrind error is fixed now (was due to an incorrect offset computation in BigInteger operator<<).
There are still some tests for edge cases in full-precision mode failing, which need to be fixed.

PR #1302, opened 2018-06-27
Guard against min/max being macros in reader.h

chwarr:
Sometimes, particularly when Microsoft's windows.h is included, min/max
are defined as macros, interfering with use of
`std::numeric_limits::min()` and the like.

To guard against this, the function name is wrapped in an extra set of
parenthesis, which inhibits function-style macro expansion.

This is a similar commit to 6e38649ec6, but fixes uses of
`std::numeric_limits` added after that commit, like those introduced in
2ea43433e2.


chwarr:
See also [related issue](https://github.com/Tencent/rapidjson/issues/1303) for a proposal I have to prevent these from cropping up in the future.

chwarr:
Looks like [the previous build failure](https://travis-ci.org/Tencent/rapidjson/builds/397483102?utm_source=github_status&utm_medium=notification) may have been transient. I've rebased this atop 129d19ba and pushed again. This will make the merge easier and trigger a new CI build.

PR #1307, opened 2018-07-02
Rename a few internal preprocessor macros to avoid naming conflicts

moretromain:
Some internal macros (COPY/TRANS/TAIL) might conflict with externally define macros with the same sketchy naming.
Prefixing them with RAPIDJSON_ might prevent such conflicts.


miloyip:
Thank you.

PR #1309, opened 2018-07-02
Updated google test to latest commit

tmiller1996:


gongminmin:
For the compiling errors in appveyor, I can repro it locally. The fix is on the way.

PR #1312, opened 2018-07-03
Detect C++11 features for Developer Studio

erikfroseth:
This patch enables various C++11 features if the code is compiled with Developer Studio compiler version 5.14 or higher.


miloyip:
Thank you

PR #1320, opened 2018-07-12
Removing always true if condition

AnomalRoil:
Since this part is in a `else if` block after a `if (value < kTen8)`, the condition `value >= kTen8` is necessarily true in that block.


miloyip:
Thank you

PR #1326, opened 2018-07-16
compilation error fix in case of max macro is included

perimeno:


perimeno:
960b9cf has full solution

PR #1328, opened 2018-07-17
Fix the unit test error with VS2017 x86 mode.

gongminmin:
The x86 binary generated by msvc 19.10+ seems wrong here. It makes the utf16le and utf16lebom unit tests failed. Add a volatile to avoid this problem.


gongminmin:
It's very likely to be a compiler bug. With or without "volatile", the binary code are in a different order.

I'll test this on the latest msvc. If it still show up, I'll file a bug against msvc.

gongminmin:
Good news is that on msvc 19.15 (VS2017 update 8), this problem no longer exist. So, what's our choice? Narrow down the volatile to (_MSC_VER >= 1910) && (_MSC_VER < 1915), or abandon this change?

gongminmin:
OK. Abandon this one. Thanks.

Issue #1336, opened 2018-07-31
PrettyWriter<FileWriterStream> on simple scalar value does not work


lelit:
Hi, I received a [bug report](https://github.com/python-rapidjson/python-rapidjson/issues/107) on python-rapidjson that I was able to easily replicate on RapidJSON itself. 

Apparently when using a `PrettyWriter<FileWriterStream>` with a single scalar value as input, nothing is emitted on output.

Consider:
```shell
$ cd example/pretty
$ make pretty
g++     pretty.cpp   -o pretty
$ echo '[10]' | ./pretty
[
    10
]
$ echo '10' | ./pretty
$
```

Thanks&bye, lele.

miloyip:
Successfully reproduced. Thank you for your report.

PR #1340, opened 2018-08-03
Wrap all WriteXxx() calls within EndValue()

lelit:
This attempts to fix issue #1336: although it seems doing the right thing, I'm not sure this is the right _fix_, as my C++ knowledge is just above 0.

Also, I tried to run the test suite, but I do not understand why I'm hitting the following:
```
$ git submodule update --init
fatal: remote error: want 0a439623f75c029912728d80cb7f1b8b48739ca4 not valid
fatal: The remote end hung up unexpectedly
Fetched in submodule path 'thirdparty/gtest', but it did not contain 0a439623f75c029912728d80cb7f1b8b48739ca4. Direct fetching of that commit failed.
```

Thank you for any advice!

### DIFF
--- a/include/rapidjson/prettywriter.h
+++ b/include/rapidjson/prettywriter.h
@@ -92,26 +92,26 @@ public:
     */
     //@{
 
-    bool Null()                 { PrettyPrefix(kNullType);   return Base::WriteNull(); }
-    bool Bool(bool b)           { PrettyPrefix(b ? kTrueType : kFalseType); return Base::WriteBool(b); }
-    bool Int(int i)             { PrettyPrefix(kNumberType); return Base::WriteInt(i); }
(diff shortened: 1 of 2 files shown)


miloyip:
Hi. I think ideally it should test all JSON types, in order to cover all the changes.

lelit:
Ok, will do. 

I had a doubt whether a the solution was right, or if instead it should be something different: would it be possible for example override _just_ the `Prefix()` method instead?

miloyip:
I think the solution is right. I forgot to modify the prettywriter while fixing writer for relaxed JSON syntax in the past.

lelit:
Ok, thank you!

miloyip:
Thank you for your contribution.

lelit:
Great, thank you!

PR #1346, opened 2018-08-15
Adding a fix to work with Embarcadero's Clang-derived compiler

SlowPokeInTexas:
Fix for Embarcadero C++ 7.x

### DIFF
--- a/include/rapidjson/rapidjson.h
+++ b/include/rapidjson/rapidjson.h
@@ -169,8 +169,12 @@
 #include "msinttypes/inttypes.h"
 #else
 // Other compilers should have this.
+// Borland's Clang derived compiler have many of these declared in stdlib.h and including
+// these causes compilation errors
+#ifndef __BORLANDC__
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


include/rapidjson/rapidjson.h
```diff
@@ -169,9 +169,13 @@
 #include "msinttypes/inttypes.h"
 #else
 // Other compilers should have this.
+// Borland's Clang derived compiler have many of these declared in stdlib.h and including
+// these causes compilation errors
```
pah: Should't we then include `<stdlib.h>` for `__BORLANDC__` instead?

SlowPokeInTexas: stdlib.h is already included (or included later, thus the compilation bug with abs()). This compiled successfully upon testing.

miloyip: Including headers of standard library should not generate error.
If this is a compiler bug, isn't that should be fixed by the compiler?

PR #1350, opened 2018-08-24
Add a target to RapidJSONConfig.cmake.in

tchernobog:
This way, users can call target_link_libraries against the imported target, which is the recommended way of doing things.

### DIFF
--- a/RapidJSONConfig.cmake.in
+++ b/RapidJSONConfig.cmake.in
@@ -13,3 +13,9 @@ get_filename_component(RapidJSON_CMAKE_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
 set( RapidJSON_INCLUDE_DIR  "@RapidJSON_INCLUDE_DIR@" )
 set( RapidJSON_INCLUDE_DIRS  "@RapidJSON_INCLUDE_DIR@" )
 message(STATUS "RapidJSON found. Headers: ${RapidJSON_INCLUDE_DIRS}")
+
+if(NOT TARGET rapidjson)
+  add_library(rapidjson INTERFACE IMPORTED)
+  set_property(TARGET rapidjson PROPERTY
+    INTERFACE_INCLUDE_DIRECTORIES ${RapidJSON_INCLUDE_DIRS})
+endif()


tchernobog:
Any hope to see this merged?

miloyip:
Can you explain why this is needed?

tchernobog:
Sure. Modern CMake is moving away from simple variables and towards providing targets. You can find the same trend in most Find*.cmake modules provided by upstream.

The reason is that variables are easy to mistype and might silently expand to nothing (target existance is instead checked at the point of usage). You also need more commands to use them, and makes for worse extensibility (you need a variable for includes, one for libraries to link against, one for compile definitions...). 

Instead with targets, you just link against a target, and all sensible properties are propagated automatically.

This makes for less code to maintain and check. See also:
https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/

zhihaoy:
Please merge something like this before releasing a new version.  It's just a new practice in CMake usage.  For example, when vcpkg installs "xsimd," after finishes, it prints
```
    find_package(xsimd CONFIG REQUIRED)
    target_link_libraries(main PRIVATE xsimd)
```
Even though xsimd is header-only.  But with rapidjson, it prints
```
    find_package(RapidJSON CONFIG REQUIRED)
    target_include_directories(main PRIVATE ${RAPIDJSON_INCLUDE_DIRS})
```
Which looks sort of outdated.

I kind of prefer the `RapidJSON::RapidJSON` target though as suggested in https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/ , as it doesn't conflict with projects that already import rapidjson target by including it somewhere in their source directory.

aphecetche:
Hi there,

This would be really really useful indeed. That way projects that use RapidJSON _and_ modern CMake techniques would not need to defined their own `FindRapidJSON.cmake` "just" to define the missing RapidJSON::RapidJSON target.

Thanks,

Kaffeine:
Hi dear @miloyip, thank you for your work!

The library is great as it is, but sadly it is missing two important things:
1. RapidJSON::RapidJSON CMake target (added in this PR).
2. A new release. :-)

Please merge this PR even if you're not going to use it — this missing bit is a must-have thing for many users.

Issue #1368, opened 2018-09-26
Parse<kParseNumbersAsStringsFlag>() raises an error on big integer


lelit:
The following slightly modified simplereader.cpp:
```cpp
#include "rapidjson/reader.h"
#include "rapidjson/error/en.h"
#include <iostream>

using namespace rapidjson;
using namespace std;

struct MyHandler {
    bool Null() { cout << "Null()" << endl; return true; }
    bool Bool(bool b) { cout << "Bool(" << boolalpha << b << ")" << endl; return true; }
    bool Int(int i) { cout << "Int(" << i << ")" << endl; return true; }
    bool Uint(unsigned u) { cout << "Uint(" << u << ")" << endl; return true; }
    bool Int64(int64_t i) { cout << "Int64(" << i << ")" << endl; return true; }
    bool Uint64(uint64_t u) { cout << "Uint64(" << u << ")" << endl; return true; }
    bool Double(double d) { cout << "Double(" << d << ")" << endl; return true; }
    bool RawNumber(const char* str, SizeType length, bool copy) { 
        cout << "Number(" << str << ", " << length << ", " << boolalpha << copy << ")" << endl;
        return true;
    }
    bool String(const char* str, SizeType length, bool copy) { 
        cout << "String(" << str << ", " << length << ", " << boolalpha << copy << ")" << endl;
        return true;
    }
    bool StartObject() { cout << "StartObject()" << endl; return true; }
    bool Key(const char* str, SizeType length, bool copy) {
        cout << "Key(" << str << ", " << length << ", " << boolalpha << copy << ")" << endl;
        return true;
    }
    bool EndObject(SizeType memberCount) { cout << "EndObject(" << memberCount << ")" << endl; return true; }
    bool StartArray() { cout << "StartArray()" << endl; return true; }
    bool EndArray(SizeType elementCount) { cout << "EndArray(" << elementCount << ")" << endl; return true; }
};

int main() {
[...]

PR #1373, opened 2018-10-08
Add test case on kParseNumbersAsStringsFlag being able to load big ints

lelit:
See issue #1368.


miloyip:
I think it should also check whether the parsed string is as expected.

lelit:
Ok, sure!

lelit:
Uhm, maybe I'm missing something, but isn't that already checked by the [handler's RawNumber() method](https://github.com/Tencent/rapidjson/blob/master/test/unittest/readertest.cpp#L1888)?

miloyip:
Oh, true.

PR #1413, opened 2018-12-01
Fix a memory leak for invalid std::regex in Schema.

ylavic:
Free the Malloc()ated memory on std::regex_error exception.


foxtacles:
I've noticed this one as well when I aas going through the source lately, haha. I'm sure whoever wrote this code was perfectly aware of the leak, and wondered if there was any reason they didn't implement it like you did here.

ylavic:
It's not always a leak actually, for instance with the default allocator(s) the memory will finally get released (precisely with the allocator itself).

PR #1414, opened 2018-12-01
Use passed in allocator for internal regex parser.

ylavic:
This both abides caller's choosen allocator and, if none is provided, avoids creating three ones (states_, ranges_ and Parse()'s).


miloyip:
Thanks

PR #1415, opened 2018-12-02
Track (sub-)schema(s) validity and provide SchemaDocument::IsValid().

ylavic:
Set valid_ = false wherever the schema is malformed (e.g. unexpected type) or incomplete (dangling "$refs"), and return it in SchemaDociment::IsValid().

### DIFF
--- a/include/rapidjson/internal/regex.h
+++ b/include/rapidjson/internal/regex.h
@@ -392,8 +392,7 @@ private:
                 }
                 return false;
 
-            default: 
-                RAPIDJSON_ASSERT(op == kOneOrMore);
+            case kOneOrMore:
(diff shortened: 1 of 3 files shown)


miloyip:
The original idea is to ignore invalid situations in schema. And user may use schema of JSON schema to validate a schema. If user ensure that the schema is valid, there is no need to validate it by code again.

ylavic:
Good point, although a bit chicken and egg... Probably json-schema-draft-4 is well-formed :)

Possibly the checks are cheap enough to be in still (actually they are mainly `else` cases)...

ylavic:
Schema validation on the schema itself makes sense, closing.

PR #1416, opened 2018-12-03
GenericRegex: don't throw/abort on syntax error (unclosed parenthesis).

ylavic:
An invalid regex shouldn't kill the process, IsValid() can be checked.


miloyip:
Thanks

PR #1425, opened 2018-12-05
Fix off by one in FileReadStream::Peek4()

ylavic:
Until Read() reaches EOF, Peek4() must not take off by one in
bufferLast_ into account; otherwise a buffer of size exactly 4 always
returns NULL.


ylavic:
Sorry for the rebase, commit message and test's buffer size were wrong..

PR #1426, opened 2018-12-10
Add "less than" operator to Pointer.

ylavic:
Allows to sort pointers in (std::)containers and/or index by them.


ylavic:
I thought it was merged, reopening...

PR #1439, opened 2019-01-10
Use the allocator of the Schema for its Pointer.

ylavic:
The Pointer passed to construct the Schema can be from the stack or any
transient storage, so the copy stored in the Schema must have the same
lifetime/allocator as the Schema itself.


miloyip:
Thank you

PR #1442, opened 2019-01-25
CMake: remove hardcoded CMAKECONFIG_INSTALL_DIR path

rbilovol:
Currently this path is hardcoded to lib/cmake.
Some distributions have different library path (like lib64).
So reuse LIB_INSTALL_DIR for that to make CMAKECONFIG_INSTALL_DIR
configurable and usable in such distros.

Signed-off-by: Ruslan Bilovol <rbilovol@cisco.com>


miloyip:
Thank you

PR #1453, opened 2019-02-10
Adding a single customization point that ensures all allocations within rapidjson can be performed with a custom memory allocator

gaspardpetit:
This change introduces the macros `RAPIDJSON_MALLOC`, `RAPIDJSON_REALLOC`, and `RAPIDJSON_FREE`;

Signed-off-by: Gaspard Petit <gaspard.petit@eidosmontreal.com>

It turns out to be quite tricky to ensure that the CRT `malloc`/`realloc`/`free` are **not** being called since the `CrtAllocator` class is used as a default template parameter in many places. In some places, the default template paramater is a `MemoryPoolAllocator` that ends up using the `CrtAllocator`.

When wanting to use a custom allocator everywhere, we end up having to redefine all the allocators, including the cases were a `MemoryPoolAllocator` is the default. This change addresses the use case of specifying a custom allocator for any/all allocations, but still wanting to use the default allocation strategies.

I considered renaming `CrtAllocator` to `DefaultAllocator`, but did not see much benefit. Let me know if you think otherwise.

I would also suggest getting rid of the `RAPIDJSON_NEW` and `RAPIDJSON_DELETE` which are not used in many places but are easy to overlook when choosing to use a custom allocator. Instead, they could use `RAPIDJSON_MALLOC` and `RAPIDJSON_FREE` by default and call the constructor/destructor explicitly.  Again, let me know if you are interested, I would not mind amending this pull-request.


eliveikis:
Any plans on merging this? I agree that a simple new/delete override that would preserve the default allocators is ideal for easy overriding.

include/rapidjson/rapidjson.h
```diff
@@ -617,6 +617,22 @@ RAPIDJSON_NAMESPACE_END
...
+#define RAPIDJSON_MALLOC(size) std::malloc(size)
+#endif
+#ifndef RAPIDJSON_REALLOC
+///! customization point for global \c realloc
+#define RAPIDJSON_REALLOC(ptr, new_size) std::realloc(ptr, new_size)
+#endif
+#ifndef RAPIDJSON_FREE
+///! customization point for global \c free
+#define RAPIDJSON_FREE(ptr) std::free(ptr)
+#endif
```
eliveikis, line 634: Better perhaps would be to expose some externally accessible hooks that could be set in code by the user of rapidjson library. By default they could point to `std::` functions and, if set by user, could point to any user-provided function.

miloyip: Using function pointers for this seems adding unnecessary overheads.
I think macro customization is sufficient for most users. And they can define a macro to call a function that can be further redirect via function pointer.

gaspardpetit: I agree, I did not see much benefit in adding function pointers. If desired, library owners can chose to define the macro such that it points to a customizable allocator.

gaspardpetit:
I have merged the latest changes, please let me know if I can do anything else to get this pull request though.

Issue #1476, opened 2019-04-03
no type named 'streamsize' in namespace 'std'


liangdzou:
➜  liangzou@LIANGZOU-MB1 ~/projects/jasmc git:(master) ✗ clang++ -I 3rdparty/rapidjson/include test.cpp                                                                      
In file included from test.cpp:4:
3rdparty/rapidjson/include/rapidjson/istreamwrapper.h:100:57: error: no type named
      'streamsize' in namespace 'std'
            if (!stream_.read(buffer_, static_cast<std::streamsize>(bufferSize_))) {
                                                   ~~~~~^
1 error generated.

This should be in header ios (https://en.cppreference.com/w/cpp/io/streamsize).

liangdzou:
add "#include <ios>" worked in my case

PR #1481, opened 2019-04-09
add missing header "ios"

liangdzou:
add missing header "ios" for the symbol "std:: streamsize"


liangdzou:
Fixed the issue #1476

PR #1485, opened 2019-04-11
Object members stored in std::multimap

ylavic:
Set RAPIDJSON_USE_MEMBERSMAP to use a (std::multi)map for object members.

When `RAPIDJSON_USE_MEMBERSMAP` is defined, an object `Value` will store
its members in an (re)allocated array of Members like before, but also
in an `std::multimap<GenericValue::Data,SizeType>` where the key and value
reference the corresponding `Member` by its Data and index in the array,
respectively, and in a relocatable manner.

The layout of the members map/array is now:
 `{multimap*}<>{capacity}<>{Member[capacity]}<>{multimap::iterator[capacity]}`
where `<>` stands for the `RAPIDJSON_ALIGN`-ment of each part, if needed.

This layout needs to be reallocated when the current capacity is
exhausted, which requires to take care of the `multimap` and its iterators
explicitely. The `multimap` is allocated separately and only its pointer is
saved in this layout, so it can easily be restored in its new position.
As for the old/alive iterators, they must move to their new offset according
to the new capacity.

With this in place, it's immediate to get the `multimap::iterator` from a
MemberIterator and vice versa, thus the same complexity applies for the
operations with MemberIterator or MapIterator.

For `FindMember()` and `RemoveMember()`, the complexity drops from `O(n)` to
the `multimap/rbtree`'s `O(log n)`.
For `EraseMember()` it drops from `O(n-m)` to `O((log n)-m)`, `m` representing
the move/copy of the trailing members.
For `AddMember()` though, the complexity grows from `O(1)` to `O(log n)` due to
the insertion in the `multimap` too.

Consequently parsing will be slower, up to ~20% measured in perftests on
my laptop (since it's mainly composed of insertions). But later work on
the `Document` (usually the goal of parsing...) will be much faster; the
new `DocumentFind` perftest included in this commit is 8 times faster with
`RAPIDJSON_USE_MEMBERSMAP` (still on my laptop). Overall the tests are 4%
slower (mainly composed of parsing), and notably 15% slower for schemas
parsing/validation (which supposedly comes from the larger JSON files
parsing, still). As a side note, when `RAPIDJSON_USE_MEMBERSMAP` is not
defined, this commit does nothing (same results for perftest with regard
to previous versions).

Finally, the `multimap` is allocated and constructed using StdAllocator,
so they will use the same Allocator than for any other `Value` allocation,
and thus will benefit from the same performance/safety/security/whatever
provided by the user given Allocator.

PR #1503, opened 2019-04-28
Fix (Sub-)Value assignment

ylavic:
Extracted from PR #1485, should probably live on its own...

The issue is, from `valuetest.cpp::MergeDuplicateKey()`:
```
        for (Value::MemberIterator itr = v.MemberBegin(); itr != v.MemberEnd(); ++itr) {
            if (itr->value.Size() == 1)
                itr->value = itr->value[0];
            ...
        }
```

Without the fix, in current `GenericValue& operator=(GenericValue&)`, `itr->value.~GenericValue()` gets called before `itr->value.RawAssign(itr->value[0])` and thus `itr->value[0]` may be used after free (with a `kneedFree` allocator).

PS: the VC2013 compilation fix for `sortkeys.cpp` is also included here to pass ci.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -916,8 +916,13 @@ public:
     */
     GenericValue& operator=(GenericValue& rhs) RAPIDJSON_NOEXCEPT {
         if (RAPIDJSON_LIKELY(this != &rhs)) {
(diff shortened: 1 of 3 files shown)


miloyip:
I suggest adding an unit test for testing this case.

ylavic:
> I suggest adding an unit test for testing this case.

Done.

PR #1557, opened 2019-08-23
Documentation: fixed Uint case

MBoldyrev:
In docs, `Uint` is sometimes spelled as `UInt`, while the actual methods and flags have small `i`. This change fixes that.


miloyip:
Thanks

PR #1592, opened 2019-10-21
Add kWriteNanAndInfAsNullFlag for writing invalid double values as null

timwoj:
This PR adds another flag to the Writer class that allows writing NaN and +/- Infinity values to the output stream as `null` instead of either outputting a blank field or using the existing option to write `NaN`/`Infinity`/`-Infinity`.


miloyip:
I think this can be done by making a writer wrapper which converts the events.

timwoj:
Probably, but given that the two ways to print these values currently produce JSON that can't be parsed by a number of parsers, I thought it'd be good to introduce this into core.

timwoj:
Any chance this is going to get merged or should I just close it out and continue using my fork?

timwoj:
Closing this as it's obviously not going to get merged.

PR #1613, opened 2019-11-24
Check before really encoding code points by default.

piratf:
Users can undef the macro 'VALIDATE_CODEPOINT' to disable this feature.

### DIFF
--- a/include/rapidjson/encodings.h
+++ b/include/rapidjson/encodings.h
@@ -691,6 +691,8 @@ struct AutoUTF {
 
     \tprarm CodePointValidation Run validate before encode or not.
 */
+// By default, This Encoder will validate code point and generate parse error.
+// Users can switch the check feature off by set 'CodePointValidation' to 'false'.
 template<bool CodePointValidation = true>
 class ValidatableEncoder {
 public:
@@ -701,7 +703,6 @@ public:
     static bool EncodeUnsafe(OutputStream &os, unsigned codepoint);
 };
 
-// By default, reader will validate code point and generate parse error.
 template<bool CodePointValidation>
 template<typename TEncoding, typename OutputStream>
 bool
@@ -724,7 +725,6 @@ ValidatableEncoder<CodePointValidation>::EncodeUnsafe(OutputStream &os, unsigned
     return true;
 }
 
-// Users can switch the check feature off by set 'CodePointValidation' to 'false'.
 template<>
 template<typename TEncoding, typename OutputStream>
 bool
@@ -733,7 +733,6 @@ ValidatableEncoder<false>::Encode(OutputStream &os, unsigned codepoint) {
     return true;
 }
 
-// Users can switch the check feature off by set 'CodePointValidation' to 'false'.
 template<>
 template<typename TEncoding, typename OutputStream>
 bool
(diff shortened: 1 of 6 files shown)


piratf:
image

It seems that the coverage decreased due to the newly added macro?

include/rapidjson/encodings.h
```diff
@@ -662,6 +694,10 @@ struct Transcoder {
         unsigned codepoint;
         if (!SourceEncoding::Decode(is, &codepoint))
             return false;
+#ifdef VALIDATE_CODEPOINT
```
miloyip: This is not ideal. Since this is a feature, I think it should be better switched on/off via non-type template parameter flag.

piratf: Good idea, I will handle this update.

include/rapidjson/document.h
```diff
@@ -2278,17 +2278,18 @@ class GenericDocument : public GenericValue<Encoding, Allocator> {
...
-        \tparam SourceEncoding Encoding of input stream
-        \tparam InputStream Type of input stream, implementing Stream concept
+        \tparam SourceEncoding Encoding of input stream.
+        \tparam CodePointValidation Enable validate or not, implemented by ValidatableEncoder.
+        \tparam InputStream Type of input stream, implementing Stream concept.
         \param is Input stream to be parsed.
         \return The document itself for fluent API.
     */
-    template <unsigned parseFlags, typename SourceEncoding, typename InputStream>
+    template <unsigned parseFlags, typename SourceEncoding, bool CodePointValidation, typename InputStream>
```
miloyip: I thought it was done via `parseFlags`...

piratf: A new flag? Is there a name recommendation?

piratf: That may introduce the cost of runtime judgment, so I took it as a second choice at first.

miloyip: Since it is a compile-time parameter, all conditions involving that parameter should be evaluated at compile-time. Actually, I have not analyzed what is the proper way to resolve this. Should `kParseValidateEncodingFlag` include this checking?

piratf: This will have two effects, the first is that the behavior after triggering a validation failure has changed from throwing an exception to returning a parse error, the second is the performance impact of more checks.
From a macro perspective, I think this behavior becomes more friendly, so I prefer to reuse the flag.
In terms of performance, we may need to observe the impact by means of flame charts, etc. If the performance impact is very obvious, we can create another mark for this check separately.
After having enough data we can explicitly declare this change in the update log and then observe the subsequent feedback.

piratf: Yes you are right about the conditions evaluation, I was thinking wrong.

miloyip:
Just some warnings to be fixed. I have no more comment.

piratf:
image

I made a red-blue differential flame chart to compare the effects before and after adding this feature. It shows that adding this feature has very little effect: 0.15% in perftest and 1.08% in unittest.

files below:
[diff_perftest.zip](https://github.com/Tencent/rapidjson/files/3976604/diff_perftest.zip)
[diff_unittest.zip](https://github.com/Tencent/rapidjson/files/3976606/diff_unittest.zip)

piratf:
Let me think about how to solve the problem of travis ci compile..

piratf:
No more errors in the code itself were found. There are still occasional problems with environment initialization.

miloyip:
Thank you for your hard work.

piratf:
It's my pleasure. Thank you for your guidance. 😃

PR #1639, opened 2020-02-10
Allow escaped apostrophe in values

xpahos:
Right now if the value has escaped apostrophe this code:

 ```
               Ch e = is.Peek();
                if ((sizeof(Ch) == 1 || unsigned(e) < 256) && RAPIDJSON_LIKELY(escape[static_cast<unsigned char>(e)])) {
                    is.Take();
                    os.Put(static_cast<typename TEncoding::Ch>(escape[static_cast<unsigned char>(e)]));
                }
```

https://github.com/Tencent/rapidjson/blob/master/include/rapidjson/reader.h#L1013

will return 0 in RAPIDJSON_LIKELY because of "\'" is not defined and later will throw an error. I think escaped apostrophe(e.g. "{"foo": "bar\'buzz"}" is not that case when we need to throw Invalid escape sequence error.


miloyip:
JSON standards do not escape apostrophe.

xpahos:
@miloyip the problem is that Java(org.json) version written by Crockford is allowed to escape apostrophe symbol https://github.com/stleary/JSON-java/blob/master/JSONTokener.java#L347. Of course, we can add something like a binary search to remove escaped sequences but a more permissive rule will be much faster(the main reason why we use rapidjson).

miloyip:
I think it is a kind of relaxed syntax (like those in #36 ) . But since it is non-standard, I prefer adding a parse flag for enabling this, if this feature would like to be merged.

xpahos:
@miloyip Unfortunately, I don't know Chinese. May be you can help me with a translation of dom.zh-cn.md for a new flag?

PR #1644, opened 2020-02-20
Closes #1643

### LINKED ISSUE #1643: Memory optimization for embedded devices
1. rapidjson uses MemoryPoolAllocator by default.
2. GenericValue uses kDefaultObjectCapacity=16 and kDefaultObjectCapacity=16.

For #1:
On embedded platforms with limited memory CrtAllocator can be a good choice. You can use CrtAllocator in a module today by explicitly creating your own typedef: 
"
typedef GenericValue<UTF8<>, CrtAllocator> Value
"
This is cumbersome when working on a platform with multiple modules using rapidjson since each module has to have the same definition of rapidjson::Value so as to interoperate.

One  solution is to create a pre-processor directive (#define RAPIDJSON_USE_CRTALLOCATOR) such that all modules that want to use CrtAllocator can do so at compile time without having to explicitly typedef them individually.

For #2: 
On the same lines as this commit: https://github.com/Tencent/rapidjson/commit/6f587466a1b714fea392423cf3507fcd0f2b546e allow default object and array capacity to be configured using  pre-processor directives.

mdamle:
This change comes up with compile time pre-processor directives to
tune the behavior of rapidjson wrt memory consumption. The idea is to
allow each module using this library to choose the right defaults based
on how it consumes memory and what performance it expects.

1. RAPIDJSON_USE_CRTALLOCATOR: If defined allows you to choose CrtAllocator over 
    MemoryPoolAllocator. If it is not defined, chooses MemoryPoolAllocator by default.
2. RAPIDJSON_VALUE_DEFAULT_OBJECT_CAPACITY and
    RAPIDJSON_VALUE_DEFAULT_ARRAY_CAPACITY: If defined and set to a
     value, uses that value for default number of objects/array elements to be pre-allocated. If not 
     defined, uses value of 16: the current default.

Verified that all tests pass.

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -56,6 +56,48 @@ class GenericValue;
 template <typename Encoding, typename Allocator, typename StackAllocator>
 class GenericDocument;
 
+/*! \def RAPIDJSON_DEFAULT_ALLOCATOR
+    \ingroup RAPIDJSON_CONFIG
+    \brief Allows to choose default allocator.
+
+    User can define this to use CrtAllocator or MemoryPoolAllocator.
+*/
+#ifndef RAPIDJSON_DEFAULT_ALLOCATOR
+#define RAPIDJSON_DEFAULT_ALLOCATOR MemoryPoolAllocator<CrtAllocator>
+#endif
+
+/*! \def RAPIDJSON_DEFAULT_STACK_ALLOCATOR
+    \ingroup RAPIDJSON_CONFIG
+    \brief Allows to choose default stack allocator for Document.
+
+    User can define this to use CrtAllocator or MemoryPoolAllocator.
+*/
+#ifndef RAPIDJSON_DEFAULT_STACK_ALLOCATOR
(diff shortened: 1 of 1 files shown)


miloyip:
Will it be more generalized to do something like `#define RAPIDJSON_DEFAULT_ALLOCATOR CrtAllocator`?

mdamle:
Yeah. That sounds better. I will make the change :)

mdamle:
This is ready for review. Please take another look :)

miloyip:
Another suggestion, I think the default allocator should be put in the template class definition, instead of the typedef:

https://github.com/Tencent/rapidjson/blob/98f52b6bb001a546968fbded3cbd6f3390bcdb41/include/rapidjson/document.h#L2153

which also reminds me that, we may also add a macro for the default stack allocator as well.

mdamle:
Makes sense. I have pushed a new commit that uses the defaults in the class template itself. I have also added anew #define for DEFAULT_STACK_ALLOCATOR.

miloyip:
Thank you

mdamle:
Thanks for all the great feedback :)

PR #1656, opened 2020-03-06
Add Option to Hasher for commutative, array element insensitivity

ClaytonSimons:
I really like the hasher here, but I have a need for it to support array element insensitivity.  I came up with a few ways to do it, but they all involve modifying Hasher.  So I came up with this edit to support it natively.  Do you think this makes sense to add to the code base?

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -221,7 +221,7 @@ class Hasher {
 public:
     typedef typename Encoding::Ch Ch;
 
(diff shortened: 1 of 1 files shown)


miloyip:
It sounds a bit weird to me. Can you ellaborate the uses of it?

ClaytonSimons:
The use for me is to allow me to use this hasher to make a method to do a quick comparison of json to tell me if the two json are equivalent regardless of order of value members or array elements.  

As I said there were some other ways I thought of to make this work such as making the stack_ member and Hash() protected so I could override EndArray() with a child class for instance.  That still requires an edit though, and so I thought maybe a more integrated approach would be appropriate?

Issue #1455, opened 2019-02-17
multiplication instead of division in StrtodNormalPrecision method


onlynishant:
Hi,

I was going through the source code and found out that `[StrtodNormalPrecision](https://github.com/Tencent/rapidjson/blob/7484e06c589873e1ed80382d262087e4fa80fb63/include/rapidjson/internal/strtod.h#L34)` uses division. I was wondering, isn't it better to use multiplication as it is generally faster than multiplication by extending `Pow10` constant map to accommodate negative pow? Is there a specific reason for that?

miloyip:
The accuracy of results should be lower if 1/(10^x) is stored as table.

PR #1668, opened 2020-03-22
extending pow10 map to incorporate negative values and using multiplication instead of division in StrtodNormalPrecision #1455

KenilMehta:


miloyip:
This does not pass unit tests as precision reduced.

KenilMehta:
Hi @miloyip 

If we look at the following code:
`#include <iostream>
#include <assert.h>
using namespace std;

int main() {
        long double pow21 = 1e+21;
        long double powNeg21 = 1e-21;
        cout.precision(308);
        long double multiplicant = 123456789;
        cout << multiplicant*powNeg21 << " " << multiplicant/pow21;
        return 0;
}
`

The values in the last line do not match exactly. Maybe that is because multiplication and division are implemented differently.
While extending the pow10 map I have checked that  (pow10(Exp) == 1/ pow10(-Exp)). 
Now as we are moving from division to multiplication, some precision loss hi happening. 
What would you suggest to me to move forward with this Issue?

PR #1677, opened 2020-03-28
Add NonConstIterator version for comparison operators

kolya7k:
Add NonConstIterator version for comparison operators to eliminate CLang 10 compile error


ostash:
If you apply this there is no need all C++20 three-way-magic-stuff. It can be just removed.

PR #1680, opened 2020-03-30
Better config module

mathstuf:
---
General improvements to the Config.cmake file.

Previous PRs:

  - #894 (@belveder79)
  - #1350 (@tchernobog)


mathstuf:
Others who may be interested (based on #1350 discussion): @zhihaoy @aphecetche @Kaffeine

mathstuf:
I suppose we can also provide the `rapidjson` target. Or the source tree could gain an alias target named `RapidJSON::RapidJSON` for consistent use for those importing the source tree.

aphecetche:
well, I would personally vote for `RapidJSON::RapidJSON` as that's what I'm using currently (https://github.com/AliceO2Group/AliceO2/blob/dev/dependencies/FindRapidJSON.cmake), _and_ because it's following the best practice of namespacing target names ;-)

Kaffeine:
Yes, RapidJSON::RapidJSON is much better and much more appropriate, because:
1. `rapidjson` can be pretty anything, but `RapidJSON::RapidJSON` can be only a target. There is a convention/recommendation to use double colons to make it obvious that the argument is a target.
2. The project name is RapidJSON and the CMake upstream convention is to use mixed case for module names. Check the list of find modules at the [upstream repo](https://github.com/Kitware/CMake/tree/master/Modules).

mathstuf:
I've send the CLA to the relevant folks on my side. I'll sign it when they approve.

mathstuf:
This fell off my plate. I've rebased and signed the CLA.

PR #1683, opened 2020-04-01
Add \v as a valid escape character (vertical tab)

ryknuth:


miloyip:
No, it is not a valid escape character in all current standards.

ryknuth:
https://en.wikipedia.org/wiki/List_of_Unicode_characters
U+000B | 11 | Vertical tab | VT

I can find it in the c++ docs, powershell docs, etc. Please point me to the current standards that you are referencing.

ryknuth:
It works in python and java. This covers the top 4 programming languages:
https://fossbytes.com/most-popular-programming-languages/

miloyip:
https://tools.ietf.org/html/rfc7159
https://www.ecma-international.org/publications/standards/Ecma-404.htm

ryknuth:
From RFC 7159:
> All Unicode characters may be placed within the
>    quotation marks, except for the characters that must be escaped:
>    quotation mark, reverse solidus, and the control characters (**U+0000
>    through U+001F**).

>  Alternatively, there are two-character sequence escape
>    representations of some popular characters.  So, for example, a
>    string containing only a single reverse solidus character may be
>    represented more compactly as "\\".

This is simply a matter of opinion. It clearly is a valid escape character as it's defined within the unicode control characters. The opinion is whether '\v' is a popular enough to be handled. Considering the top 5 used languages support '\v', it seems it would be beneficial to support given that the library is for C++.

miloyip:
JSON is not related to popular programming languages. It is an data interexchange format.

From RFC 7159:
~~~
char = unescaped /
          escape (
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
              %x75 4HEXDIG )  ; uXXXX                U+XXXX
~~~

It is possible to support relaxed syntax, as RapidJSON does support several non-standard syntax via additional options. But by default it must be rigid to the standard.

PR #1686, opened 2020-04-06
Fix escape solidus character ("/") in Writer

cyberbobs:
According to [RFC 7159 Section 7](https://tools.ietf.org/html/rfc7159#section-7) solidus character (`/`) must be escaped by replacing it with `\/`. Reader's [ParseStringToStream](https://github.com/Tencent/rapidjson/blob/master/include/rapidjson/reader.h#L993) method handles it standard way but escape character map in Writer was not escaping `/` at all.

This commit fixes the behavior and its test not complying to RFC 7159.


miloyip:
No.

In the first paragraph:

~~~
All Unicode characters may be placed within the
   quotation marks, **except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).**
~~~

Escaping solidus is not a MUST. And it is very often that JSON contains unescaped solidus, such as URLs as string.

cyberbobs:
Agreed. Sorry for wasting your time.

PR #1693, opened 2020-04-15
Use the found ccache when one is found, rather than "ccache" which mi…

Romain-Geissler-1A:
…ght not be in the $PATH.

I will test this in my own build, for now it's failing with an error saying that raw "ccache" executable is not found.


Romain-Geissler-1A:
Ping

PR #1704, opened 2020-04-30
Improving comparison implementation.

BRevzin:


BRevzin:
> Pending — Contributor License Agreement is not signed yet. 

What am I supposed to do here?

BRevzin:
Uh... ping?

include/rapidjson/document.h
```diff
@@ -250,15 +253,15 @@ class GenericMemberIterator {
...
-    template <bool Const_> bool operator<=(const GenericMemberIterator<Const_, Encoding, Allocator>& that) const { return ptr_ <= that.ptr_; }
-    template <bool Const_> bool operator>=(const GenericMemberIterator<Const_, Encoding, Allocator>& that) const { return ptr_ >= that.ptr_; }
-    template <bool Const_> bool operator< (const GenericMemberIterator<Const_, Encoding, Allocator>& that) const { return ptr_ < that.ptr_; }
-    template <bool Const_> bool operator> (const GenericMemberIterator<Const_, Encoding, Allocator>& that) const { return ptr_ > that.ptr_; }
-
 #ifdef __cpp_lib_three_way_comparison
-    template <bool Const_> std::strong_ordering operator<=>(const GenericMemberIterator<Const_, Encoding, Allocator>& that) const { return ptr_ <=> that.ptr_; }
+    friend auto operator<=>(Iterator, Iterator) = default;
+#else
+    bool operator==(ConstIterator that) const { return ptr_ == that.ptr_; }
```
miloyip, line 259: Should be Iterator?

BRevzin: Given:
```cpp
Iterator i;
ConstIterator ci;
```
we need the following to all work:
```cpp
i == i;
i == ci;
ci == i;
ci == ci;
```
With C++20, the defaulted `<=>` is good enough because both the `i == ci` and `ci == i` cases invoke the `ConstIterator`'s `operator==` and convert the `Iterator` into a `ConstIterator`. This is the only candidate, 

With C++17, if we take a `ConstIterator` on the right-hand-side, when which iterator we get can be convertible to that, and it doesn't matter what we have on the left-hand side. 

Demo for multiple compiler versions: [compiler explorer](https://godbolt.org/z/fMYqK4)

PR #1707, opened 2020-04-30
Disable valgrind unittest for Mac OS

woodroof:
There is no valgrind available for modern Mac OS :(


Arfrever:
Hardcoding another platform not supported by Valgrind is bad idea.
Better, generic patch in #1808.

ClausKlein:
IMHO it may be better to use `find_program(VALGRIND_FOUND valgrind)`
see too https://github.com/Tencent/rapidjson/pull/1810

PR #1568, opened 2019-09-12
Make GenericMemberIterator::Iterator public again (RAPIDJSON_NOMEMBER…

ericrannaud:
…ITERATORCLASS)

d87b698d0f made all definitions of GenericMemberIterator consistent as
classes (they were structs with RAPIDJSON_NOMEMBERITERATORCLASS
defined), but it didn't keep the member definitions public.

document.h:586:71: error: 'Iterator' is a private member of
      'rapidjson::GenericMemberIterator<false, rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >'
    typedef typename GenericMemberIterator<false,Encoding,Allocator>::Iterator MemberIterator;  //!< Member iterator for i...
                                                                      ^
document.h:2124:32: note: in instantiation of template class
      'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >' requested here
class GenericDocument : public GenericValue<Encoding, Allocator> {

PR #1806, opened 2020-12-18
Fixes for CCI build

SeleDreams:
to be able to build CCI, these two properties had to be made public


markfoodyburton:
See also https://github.com/Tencent/rapidjson/pull/1568

miloyip:
Merged #1568

PR #1817, opened 2020-12-27
Fix #1721

### LINKED ISSUE #1721: C++20 Warning: operator '|': deprecated between enumerations of different types
Hello,

I've been using the latest version (from vcpkg) of RapidJSON with a project of mine that uses C++20. When compiled under MSVC using `/W4`, the compiler generates multiple warnings that says:
```
rapidjson\document.h(1924): warning C5054: operator '|': deprecated between enumerations of different types
rapidjson\document.h(1925): warning C5054: operator '|': deprecated between enumerations of different types
...
```
For me, this is the only place where the compiler gives this warning. It seems that simply `static_cast`ing to `int` solves the issue, but I'm not sure about the compatibility of this solution.

lukedan:
Fix #1721. I also deleted the copy constructor and copy assignment of `SchemaDocumentProvider` in `test/unittest/schematest.cpp` because they don't have function bodies. I'm not sure if this'll create issues for other platforms, but for MSVC having those definition causes problems.


lukedan:
Deleting the constructors causes warnings with g++ which are treated as errors, so I'm adding empty bodies to the constructors/assignment operators.

miloyip:
Thank you

Issue #1404, opened 2018-11-21
Transform GenericArray and GenericObject to GenericValue


slsyy:
I would like to resolve JSON Pointer on GenericObject or GenericArray. Unfortunately it seems like i can't get GenericValue for mentioned types, because ValueType is private and there is no getter. Did I omit something or maybe It's not possible?

miloyip:
Yes, currently it does not provide API for this purpose.

slsyy:
I can post PR.  Would an implicit conversion operator be ok?
```cpp
operator ValueType&() const  { return value_;}
```

miloyip:
I cannot imagine any issue at the moment.
Can you try to add this and a unit test (for resolving JSON pointer) in a PR?

PR #1821, opened 2021-01-05
Add implicit conversion from Object and Array to Value (#1404)

slsyy:
Allows resolution of JSON Pointer on Object and Array


miloyip:
thank you

Issue #1448, opened 2019-02-08
no member as GetObject


emnkoksal:
I want to get Object just like in tutorial:

http://rapidjson.org/md_doc_tutorial.html

document.GetObject()

However it gives me this error:

no member named GetObjectA in rapidjson::GenericDocument<rapidjson::UTF8<>> did you mean GetObject

I cannot use GetObject() with neither rapidjson::Document, nor rapidjson::Value, etc.

How could I use GetObject()?

jcmonnin:
This is most likely because of the `GetObject` macro defined in the windows headers (in wingdi.h):

    WINGDIAPI int   WINAPI GetObjectA(__in HANDLE h, __in int c, __out_bcount_opt(c) LPVOID pv);
    WINGDIAPI int   WINAPI GetObjectW(__in HANDLE h, __in int c, __out_bcount_opt(c) LPVOID pv);
    #ifdef UNICODE
    #define GetObject  GetObjectW
    #else
    #define GetObject  GetObjectA
    #endif // !UNICODE

Long time ago some genius at microsoft thought it was a good ideal to redefine the meaning of `GetObject` in the whole world. When "windows.h" is included systematically, there is no problem as the rapidjson `GetObject` gets translated to `GetObjectA` or `GetObjectW` everywhere. When you have "windows.h" in some places, then the trouble start. In my code I have to do some workaround to make sure I never include "windows.h" together with rapidjson. When I had that issue, I was surprised I couldn't find much about this online. I guess this must be a rather common problem.

firewave:
I was able to work around this issue by adding the following after all includes:

```
#ifdef _MSC_VER
#undef GetObject
#endif
```

smeualex:
I faced now the same issue.
I'm including <windows.h> in my logger class and needed a log when parsing a json. Long story short, my build was full of errors generated by GetLogger()

My workaround is dubious at best: I reversed the order of the includes:
```
#include "Logger/Logger.h"       // ----> this includes windows.h
#include "rapidjson/document.h"
```

This is not ok, because I hardly fixed anything and it depends on inclusion order; not to mention I'm not 100% sure why it works.

Unfortunately for the moment it's not an option to remove Windows.h from the Logger.

VikingExplorer:
I know folks are trying to make a point about the stupidity of windows.h.  You're right.

However, stubborn is as stubborn does.  All the authors of this library have to do is to rename their function to get_object

ElJaviLuki:
[...]

PR #1831, opened 2021-01-20
Fix Conflict between rapidjson::GenericValue<···>::GetObject() vs GetObject macro in the windows headers (wingdi.h)

ElJaviLuki:
Fix Function Name Conflict between rapidjson::GenericValue<···>::GetObject() and GetObject macro defined in the windows headers (wingdi.h)

### DIFF
--- a/example/archiver/archiver.cpp
+++ b/example/archiver/archiver.cpp
@@ -9,8 +9,8 @@ using namespace rapidjson;
 
 struct JsonReaderStackItem {
     enum State {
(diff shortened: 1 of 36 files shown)


miloyip:
This will break all existing user code.

hendrikmuhs:
I proposed an alternative in https://github.com/Tencent/rapidjson/issues/1448#issuecomment-819285668

I happily create a PR if this gets positive feedback.

Issue #1835, opened 2021-01-28
Allow all validation failures to be reported


smhdfdl:
My requirement is to parse and validate a JSON document received from a remote source against a JSON schema, and to report all the validation failures in the document.  In other words, do not quit when the first validation failure is encountered, but instead add all failures to the validator's error structure.  The software that will use RapidJSON already does this for XML documents that are validated with XML Schema, and text/binary documents that are parsed and validated with DFDL schema.

smhdfdl:
I will shortly be submitting a PR to provide this feature.

smhdfdl:
PR merged

Issue #1836, opened 2021-01-28
Provide human-readable validation failure messages


smhdfdl:
My requirement is to parse a JSON document received from a remote source against a JSON schema, and to report validation failures in a human-readable way.  RapidJSON already provides such messages for parse errors, it would be nice to do the same for validation failures too.

smhdfdl:
I will shortly be submitting a PR to provide this feature.

Issue #1805, opened 2020-12-11
Access violation on Windows only when Append() to PointerType() with own allocator


smhdfdl:
I have made a prototype change to `schema.h` in my fork so that property dependency errors appear in the same format as schema dependency errors. (See issue #1835 and PR #1837). This required that I create a new `schemaRef` pointer that appends `/dependencies/{sourceName}` to the invalid schema pointer in order to correctly reflect the schema location. Unit tests updated to expect the new error format. The code to do this is in `EndMissingDependentProperties()`. Specifically, note the two lines below where I get an existing PointerType created earlier by a Schema and append to it, passing an allocator `<alloc>`.

```
    void EndMissingDependentProperties(const SValue& sourceName) {
        if (!missingDependents_.Empty()) {
            // Create equivalent 'required' error
            ValueType error(kObjectType);
            error.AddMember(GetMissingString(), missingDependents_.Move(), GetStateAllocator());
            AddErrorInstanceLocation(error, false);
            PointerType schemaRef = GetInvalidSchemaPointer().Append(SchemaType::GetDependenciesString(), <alloc>);
            AddErrorSchemaLocation(error, schemaRef.Append(sourceName.GetString(), sourceName.GetStringLength(), <alloc>));
            ValueType wrapper(kObjectType);
            wrapper.AddMember(ValueType(SchemaType::GetRequiredString(), GetStateAllocator()).Move(), error, GetStateAllocator());
            currentError_.AddMember(ValueType(sourceName, GetStateAllocator()).Move(), wrapper, GetStateAllocator());
        }
    }
```

1) If `<alloc>` is the validator's allocator `&GetStateAllocator()` all unit tests pass but I need to change `perftest/schematest.cpp` to use a templated SchemaDocumentType and the `Schema.TestSuite` perf test fails with an assertion error. 
2) If I do **not** pass an allocator, all unit tests pass on Linux & Mac, but the `TestSuite` unit test fails on Windows, specifically failing during test `jsonschema/tests/draft4/dependencies.json "multiple dependencies" "missing dependency"`. The failure is `SEH exception with code 0xc0000005 thrown in the test body`, an access violation. 
3) If `<alloc>` is the pointer's allocator obtained via its `&GetInvalidSchemaPointer().GetAllocator()` method then all unit & perf tests pass.

The code has let me do 1) and 2) but it's not clear to me whether this should be allowed and there is a bug, or it should not be allowed.

smhdfdl:
[...]

Issue #1648, opened 2020-02-23
JSON Schema draft v6 and v7 or maybe v8 support


mohsenomidi:
Hi to all, 
Greetings!

Are there any plans to support drafts later than 4? ,For example [Draft 7](http://json-schema.org/draft-07/json-schema-release-notes.html)  or [2019-09 (formerly known as Draft 8)](http://json-schema.org/specification-links.html#2019-09-formerly-known-as-draft-8).

Regards,

DubbeleJ:
Is there any intention or is there anyone working on extending the schema validation support as suggested in this Issue?

eliveikis:
Draft 7 support needed here.

sam20908:
Needed here too.

eefan000:
+1

piggyy:
Draft v6 support needed

PR #1837, opened 2021-01-28
Fixes for issues #1835 & #1836 - Multiple validation failures and readable validation messages

smhdfdl:
This PR adds code and unit tests that fix issue #1835 and issue #1836. Apologies for the changes being together but they were made at the same time.

**1836 Readable validation messages**
- In `error.h` define an enum `ValidateErrorCode` for a set of `kValidateErrorXXX` error codes, one per validation failure type
- In `en.h` add `GetValidateError_En()` to map error code to a human-readable error message with inserts
- In `schema.h`:
  - When creating a validation error, use the error code as the primary indicator, in addition to the error keyword
  - Add `GetInvalidSchemaCode()` to validator to return the error code
  - `oneOf`: Split into two errors - 'no match' and 'multiple match' - with error code and message for each
  - `allOf`: Create in accordance to https://rapidjson.org/md_doc_schema.html#allOf-anyOf-oneOf 
  - `dependencies`: Make property dependency error look like schema dependency error (bug found - see issue #1805)
  - `additionalItems`: Correct error keyword from `items` to `additionalItems`
  - Refactor `AddErrorLocation()` 
- In `schematest.cpp` fix the expected test outputs to include error code 
- In `schemaValidator.cpp` example, add code to walk the errors and output error messages with inserts 

**1835 Multiple validation failures**
- In `schema.h`:
  - Add `ValidateFlag` enum and member `flags_` to the validator (analogous to `ParserFlag` for the reader)
  - Provide a flag `kValidateContinueOnErrorFlag`
  - Provide `GetValidateFlags()` and `SetValidateFlags()` for access to `flags_` member
  - If `kValidateContinueOnErrorFlag` set, do not set the `valid_` member and thus allow the validator to continue
  - Do **not** change macro `RAPIDJSON_INVALID_KEYWORD_RETURN` - it still always returns false (see note)
  - Change `IsValid()` to return false if `kValidateContinueOnErrorFlag` set and the `error_` object is not empty
  - Change `GetInvalidSchemaKeyword()` and `GetInvalidSchemaCode()` similarly
  - At various places ensure that internal variables are in a correct state, so that continuing after an error is safe
  - Allow sub-validators to inherit the setting of `kValidateContinueOnErrorFlag`
  - Force sub-validators for `oneOf`, `anyOf`, `allOf`, `not`, `dependencies` to **not** inherit `kValidateContinueOnErrorFlag` 
  - Add `ResetError()` to allow errors to be reset (useful for incremental parsing)
- In `schematest.h`:
  -  Add tests to exercise the validator when `kValidateContinueOnErrorFlag` set
  - Add tests to specifically exercise that validator internal variables are correctly set after an error 
  - Change `VALIDATE` and `INVALIDATE` macros 
- In `bin` folder add new `unittestschema` folder containing new schemas for tests

Notes:
1) To allow continuing after an error, I considered changing macro `RAPIDJSON_INVALID_KEYWORD_RETURN` so that it did not return `false` if `kValidateContinueOnErrorFlag` set, but instead continued. I coded a prototype, but it meant that unnecessary and unsafe processing continued (eg, creating sub-validators when type wrong).
2) The change to `allOf` error reporting. This was also made to aid the user in locating the failure, but is a change in behaviour. We could control this via a validation flag if necessary.


include/rapidjson/schema.h
```diff
@@ -113,13 +114,36 @@ inline void PrintValidatorPointers(unsigned depth, const wchar_t* s, const wchar
...
+    User can define this as any \c ValidateFlag combinations.
+*/
+#ifndef RAPIDJSON_VALIDATE_DEFAULT_FLAGS
+#define RAPIDJSON_VALIDATE_DEFAULT_FLAGS kValidateNoFlags
+#endif
+
+//! Combination of validate flags
+/*! \see
+ */
+enum ValidateFlag : unsigned {
```
miloyip, line 141: This syntax is C++11 only.

example/schemavalidator/schemavalidator.cpp
```diff
@@ -7,9 +7,124 @@
...
 using namespace rapidjson;
 
+typedef GenericValue<UTF8<>, CrtAllocator > ValueType;
+
+// Forward ref
+static void CreateErrorMessages(const ValueType& errors, size_t depth, const char* context);
+
+// Convert GenericValue to std::string
+static std::string GetString(const ValueType& val) {
+    std::string str("");
```
miloyip, line 22: How about just use Writer to output a JSON string for this conversion?

miloyip: `to_string()` is C++11 as well.

miloyip:
This PR looks promising.
Please fix errors indicated by appveyor and travis-ci.

include/rapidjson/schema.h
smhdfdl: Fixed

example/schemavalidator/schemavalidator.cpp
smhdfdl: @miloyip I didn't use Writer because this is not for writing JSON value. I just want the stringified value, for use as message insert. Writer will add double quotes for String type. Please show how you would prefer to fix this.

smhdfdl: Writer outputs JSON values, so strings are quoted, which is not what I want here. I have used ostringstream instead. (It would be nice if GenericValue had a to_string() method :) )

smhdfdl:
@miloyip One of the Travis builds is failing.  I believe this is hitting the same problem as issue #1089. Note the output from the 2nd test - the `SchemaValidator.TestSuite` is failing the `multipleOf` test, but because the `refRemote` test is expected to fail, the `multipleOf` test does **not** cause the overall `TestSuite` test to fail. I think this failure is always present.  In my new  `ContineOnErrors` test, I have added a `multipleOf failure with a value of '0.1`. This fails as explained in #1089. Please advise what we should do here. I could change my test, but it is worrying that this problem is present in the code.

```
[ RUN      ] SchemaValidator.TestSuite
Fail: jsonschema/tests/draft4/multipleOf.json "by small number" "0.0075 is multiple of 0.0001"
Fail: jsonschema/tests/draft4/refRemote.json "change resolution scope" "changed scope ref invalid"
 261 / 263 passed (99%)
 [       OK ] SchemaValidator.TestSuite (2 ms)
...
 [ RUN      ] SchemaValidator.ContinueOnErrors
 /home/travis/build/Tencent/rapidjson/test/unittest/schematest.cpp:2293: Failure
 Value of: true == d.Accept(validator)
   Actual: false
 Expected: true
 /home/travis/build/Tencent/rapidjson/test/unittest/schematest.cpp:2293: Failure
 Value of: true == validator.IsValid()
   Actual: false
 Expected: true
 /home/travis/build/Tencent/rapidjson/test/unittest/schematest.cpp:2293: Failure
 Value of: code == kValidateErrorNone
   Actual: false
 Expected: true
 /home/travis/build/Tencent/rapidjson/test/unittest/schematest.cpp:2293: Failure
 Value of: validator.GetInvalidSchemaKeyword() == 0
   Actual: false
 Expected: true
 Invalid schema: #/definitions/decimal_type
 Invalid keyword: multipleOf
 Invalid code: 1
 Invalid message: Number '%actual' is not a multiple of the 'multipleOf' value '%expected'.
 Invalid document: #/version
 Validation error: {"multipleOf":{"actual":1.0,"expected":0.1,"errorCode":1,"instanceRef":"#/version","schemaRef":"#/definitions/decimal_type"}}
 [  FAILED  ] SchemaValidator.ContinueOnErrors (0 ms)
```

miloyip:
I think that the floating point issue should not be tackled in this PR.
I suggest changing your unit test for now.

smhdfdl:
@miloyip Unit test fixed, all checks pass.

Please let me know what the outlook is for the PR being merged. I have some other changes in the pipeline, specifically to address issue #1648 and further ahead #1486.

miloyip:
Thank you for your great effort 🍺

foxtacles:
Thank you for this valuable contribution @smhdfdl!

I may have found an issue with the continuation of the schema validation during an experiment of mine:

```
rapidjson::Document schema;
rapidjson::Document input;

schema.Parse("{\"type\":\"string\", \"anyOf\":[{\"maxLength\":2}]}");
input.Parse("3");

auto schema_document = rapidjson::SchemaDocument(schema);
rapidjson::SchemaValidator validator(schema_document);
validator.SetValidateFlags(rapidjson::RAPIDJSON_VALIDATE_DEFAULT_FLAGS | rapidjson::kValidateContinueOnErrorFlag);
input.Accept(validator);
```

This program crashes when setting the `kValidateContinueOnErrorFlag`. It completes successfully when the flag is not set. It also completes successfully (with or without the flag) if the `anyOf` array in the schema is empty. I don't have more time right now to investigate, but maybe you are able to reproduce the issue and provide a fix. Thank you!

smhdfdl:
@foxtacles Thanks for trying out the updated code, and for finding the crash. It is a bug. Fixed in PR #1844.

PR #1844, opened 2021-02-12
After PR 1837, fix crash where simple type with sub-schema has a bad value

smhdfdl:
See comment from @foxtacles here https://github.com/Tencent/rapidjson/pull/1837#issuecomment-777717462.

Scenario is a schema fragment with sub-schemas (eg, `anyOf`). If such a schema is flagged with an error (eg, type wrong) then the sub-validators for those schemas will not get created because `Schema::CreateParallelValidators()` is not called. When the validator is subsequently called for its `EndValue()` event, and the continue-on-error flag is set, it in turn calls `Schema::EndValue()` but this does not check that it has sub-validators before calling `context.validators[i]->IsValid()`. This causes the crash described in https://github.com/Tencent/rapidjson/pull/1837#issuecomment-777717462.

This PR fixes the code by checking that the validator count is greater than zero, and adds a unit test.

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -733,6 +733,7 @@ public:
     }
 
     RAPIDJSON_FORCEINLINE bool EndValue(Context& context) const {
+        // Only check pattern properties if we have validators
         if (context.patternPropertiesValidatorCount > 0) {
             bool otherValid = false;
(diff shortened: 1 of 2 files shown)


smhdfdl:
Fyi @miloyip

Issue #1846, opened 2021-02-22
[C++20] Infinite loop on GenericValue::operator==


stac47:
Hello,

Building the following program with C++20 standard leads SIGSEGV.
```c++
#define RAPIDJSON_HAS_STDSTRING 1
#include <rapidjson/document.h>
#include <rapidjson/pointer.h>

int main(int argc, const char *argv[])
{
    std::string jsonStr = "{\"a\":1}";
    rapidjson::Document doc;
    doc.Parse(jsonStr);
    const rapidjson::Value& a = doc["a"];
    (void) (a == 1);
    return 0;
}
```
A warning is emited by GCC `gcc (GCC) 11.0.0`:
```
/rapidjson/document.h:1089:163: warning: in C++20 this comparison calls the current function recursively with reversed arguments
 1089 |     template <typename T> friend RAPIDJSON_DISABLEIF_RETURN((internal::IsGenericValue<T>), (bool)) operator==(const T& lhs, const GenericValue& rhs) { return rhs == lhs; }
```
Regards,

PR #1847, opened 2021-02-22
Fix recursive operator== call in C++20 (#1846)

stac47:
Hello,
I propose this fix for #1846. I validated it works in C++20 with gcc 11 and causes no regression in the tests suite.
Regards,


miloyip:
Thank you

Issue #1843, opened 2021-02-10
Implement 'id' keyword when validating


smhdfdl:
As per https://rapidjson.org/md_doc_schema.html#Conformance, one of the Draft-04 unit tests is failing because the `id` keyword is not processed when compiling a JSON schema.  I have made changes to support this fully, in accordance with https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01#section-8. 

Note that although that is the Draft-07 specification, the Draft-04 specification (https://tools.ietf.org/html/draft-zyp-json-schema-04#section-7) did not fully specify the behaviour, leading to uncertainty about how it should be implemented. This was clarified by Draft-07, and not listed under the differences between Draft-07 and earlier drafts, the implication being that is the correct behaviour for earlier drafts also. (The one stated difference is that in Draft-06 and later, `id` is renamed `$id`).

Several new unit tests added to exercise the multitude of possibilities. The one failing Draft-04 unit test now passes.
 
See PR #1848.

Noted that `GenericSchemaDocument` could detect some error scenarios and create an error structure like `GenericSchemaValidator`. This is addressed by issue #1841.

smhdfdl:
@miloyip Your thoughts are welcome please.

PR #1848, opened 2021-02-25
Fix issue 1843 - support Id keyword

smhdfdl:
See issue #1843.  Note that this PR also includes PR #1393 as it affects the same area of code, and I have already merged #1393 into my fork.

Also note that `URIType` in `GenericSchemaDocument` has been replaced by `SValue` to avoid confusion with the new `Uri` class, and to match `Schema`.


smhdfdl:
@miloyip As this PR completes the code needed to pass 100% of the Draft 4 test suite, I have uncommented the check in `schenmatest.cpp` test `TestSuite` that passes = total. However one of the architecture/compiler combinations in the Travis CI build fails because of this, with a totally unrelated error, which looks like a separate bug.

example/schemavalidator/schemavalidator.cpp
```diff
@@ -2,6 +2,8 @@
 
 // The example validates JSON text from stdin with a JSON schema specified in the argument.
 
+#define RAPIDJSON_HAS_STDSTRING 1
```
miloyip, line 5: This seems unncessary

include/rapidjson/schema.h
```diff
@@ -1552,6 +1592,209 @@ struct TokenHelper<Stack, char> {
...
+///////////////////////////////////////////////////////////////////////////////
+// Uri
+
+template <typename SchemaDocumentType>
+class Uri {
+public:
+    typedef typename SchemaDocumentType::Ch Ch;
+    typedef typename SchemaDocumentType::AllocatorType AllocatorType;
+    typedef internal::Schema<SchemaDocumentType> SchemaType;
+    typedef std::basic_string<Ch> String;
```
miloyip, line 1604: I remembered rapidjson was not using std::string internally elsewhere. It is due to some legacy issues.

include/rapidjson/schema.h
```diff
@@ -1552,6 +1592,209 @@ struct TokenHelper<Stack, char> {
 
 } // namespace internal
 
+///////////////////////////////////////////////////////////////////////////////
```
miloyip, line 1595: If this can be used out of schema, I think we may put it in separated file, similar to Pointer, regex. And then Schema depends on Uri and not vice versa. I have not read the schema yet. If URI is not part of public API, we may put it in internal folder.

include/rapidjson/schema.h
smhdfdl: Needed to stop compiler errors

example/schemavalidator/schemavalidator.cpp
smhdfdl: Needed to stop compiler errors

include/rapidjson/schema.h
smhdfdl: Yes I did consider that, but at the moment it uses the `GetXxxxString()` functions of `Schema` for convenience. It is exposed externally .... we have added a second flavour of resolver call, so that the resolver gets the Uri object, for convenience.

smhdfdl:
There is more work to do on this PR.  As coded, RapidJSON assumes that any $ref that does not start with a '#' is remote. That is not true. The algorithm should be that a $ref is resolved, then a search is made first for that resolved reference in the local schema. This behaviour has been clarified in later drafts, but I believe that draft 4 should also behave like this (see https://jsonschemalint.com/#!/version/draft-04/markup/json). Further, JSON pointers are relative to their base URI, not the root of the schema.  This is a good example to show all this:
```
{
       "id": "http://example.com/root.json",
       "definitions": {
           "A": { 
               "id": "#foo",
               "type": "integer"
           },
           "B": {
               "id": "other.json",
               "definitions": {
                   "X": { 
                       "id": "#bar",
                       "type": "boolean"
                   },
                   "Y": {
                       "$ref": "#/definitions/X"
                   }
               }
           }
       },
       "properties": {
           "PX1": {
               "$ref": "#/definitions/B/definitions/Y"
           },
           "PX2": {
               "$ref": "other.json#/definitions/X"
           },
           "PX3": {
               "$ref": "other.json#bar"
           },
           "PA1": {
               "$ref": "#/definitions/A"
           },
           "PA2": {
               "$ref": "#foo"
           }
       }
   }
```

include/rapidjson/schema.h
ylavic: The issue with std::string is that it requires an allocation when all you have is a ptr+len tuple (i.e. std::string_view or rapidjson's [Generic]StringRef).

IIUC, all RAPIDJSON_HAS_STDSTRING constructs with const sd::string& allow for passing them in many places, but any internal value.GetString() or StringRef() => std::string conversion could be costly.

Maybe, rather, more handling of std::basic_string_view is needed in rapidjson, to replace GenerericStringRef (progressively?), since a string_view can be constructed cheeply from both a std::string and a GenericStringRef, and since a string_view has the same interface as std::string it's may be more welcoming to contributors..

ylavic: At least if std:string is used as here to store allocated data, it'd need to take an user provided Allocator as argument (IMHO)..

smhdfdl:
@miloyip Please can you help with this question, it's blocking my progress.  In `GenericSchemaDocument::CreateSchemaRecursive`, when it calls itself recursively why is the `schema` argument set to `0` ?
```
CreateSchemaRecursive(0, pointer.Append(itr->name, allocator_), itr->value, document)
```

smhdfdl:
@ylavic  You may have noted that your useful PR #1393 is included in this PR, as this code area needs changes.

smhdfdl:
As the PR stands, and because it includes PR #1393, when resolving a local `$ref` in `HandleRefSchema()`, the JSON pointer fragment is used to locate the corresponding place in the parsed JSON schema document using `pointer.Get(document)`. That now won't always work because JSON pointers are relative to the current base URI, which gets changed by the presence of `id` properties in the schema. So we need to establish `base` and then use `pointer.Get(base)`. Two ways to obtain `base`. 1) Search the schema for the resolved `id` that matches the resolved `$ref` (ignoring fragment), 2) Pre-scan the document for `id` and build a map of resolved `id`s which can then be searched. Option 2) avoids multiple searches, but Option 1) forces a scan when there might not be any `id`s (other than top-level). @miloyip @ylavic any thoughts?

ylavic:
@smhdfdl I wonder if SchemaDocument would better have two distinct methods, ::Parse() and ::Compile().

The ::Parse() method would be using Document::Parse() with a JSON schema specific filter that normalizes the "$ref"s of the resulting document, and builds a catalog.
The SAX-like filter maintains the current Pointer/tokens of the element being visited, and when it finds an "id" maintains the current absolute Uri (probably in a stack).
So each time the filter finds a relative "$ref" it can substitute the string with its absolute form.
Also, each time the filter finds a new absolute "id" it can associate the Uri with the current Pointer, thus building a catalog (map of absolute "id" => root relative Pointer).

Now when the ::Compile() method finds a "$ref" it is either:
1. relative (to the root) => usual/simple pointer lookup applies;
2. absolute and prefixed by something in the catalog => use the associated pointer + suffix to resolve;
3. absolute and not prefixed by something in the catalog => let the RemoteSchemaDocumentProvider find it.

That's two passes overall for parsing and compiling, instead of one pass for parsing and either two passes for compiling or the brute force.

What do you think?

smhdfdl:
@ylavic Yes, I had been thinking that a pre-walk of the schema document would be more efficient for schemas that use several ids and/or $refs from within the `definitions` section. I think your cases 1 & 2 collapse together, as the root document has a URI too. There are two cases to handle, JSON pointer fragments and plain name fragments.

I have code working based on my earlier approach, it handles all the cases in the example schema in the comment above. I need to add caching of all resolved ids, both local and remote, and make some other improvements.

include/rapidjson/schema.h
smhdfdl: Latest commit renames `Uri` to `GenericUri` and moves it to new file `uri.h`

smhdfdl:
@miloyip There are Travis failures, has something changed recently?
```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install cmake valgrind clang-8
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have requested an impossible situation or if you are using the unstable distribution that some required packages have not yet been created or been moved out of Incoming. The following information may help to resolve the situation:
The following packages have unmet dependencies:
 clang-8 : Depends: libclang-common-8-dev (= 1:8-3~ubuntu16.04.1) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
apt-get.diagnostics
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install cmake valgrind clang-8" failed and exited with 100 during .
```

miloyip:
> @miloyip There are Travis failures, has something changed recently?
I didn't. The yaml in master seems not changed for a year. May have travis changed something?

smhdfdl:
@miloyip @ylavic I am keen to get this PR merged as I have other changes backing up behind it. 

Travis now seems to be working again. There is a failure see this earlier comment: https://github.com/Tencent/rapidjson/pull/1848#issuecomment-786319681.

What is the easiest way to fix the fact that I am using std::string in the new Uri class? I am not a deep C++ programmer. If one of you can assist I would appreciate it.

Coveralls is reporting -0.008%, I will look at that.

test/unittest/pointertest.cpp
```diff
@@ -1,6 +1,7 @@
 // Tencent is pleased to support the open source community by making RapidJSON available.
 // 
 // Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip.
+// Portions (C) Copyright IBM Corporation 2021
```
miloyip, line 4: I am not sure if this is legally OK for my company. Can it be removed?

include/rapidjson/uri.h
```diff
@@ -0,0 +1,257 @@
...
+                start = pos2;
+            }
+        }
+        // Look for fragment (#(.*))?
+        if (start < len) {
+            frag_ = uri.substr(start);
+        }
+        base_ = scheme_ + auth_ + path_ + query_;
+        uri_ = base_ + frag_;
+        //std::cout << " Parsed uri: " <<  "s: " << scheme_.c_str() << " a: " << auth_.c_str() << " p: " << path_.c_str() << " q: " << query_.c_str() << " f: " << frag_.c_str() << std::endl;
```
miloyip, line 199: Please remove these debug code

miloyip:
> What is the easiest way to fix the fact that I am using std::string in the new Uri class? I am not a deep C++ programmer. If one of you can assist I would appreciate it.

I think this will be better. I am not sure if I can help you to rewrite that part recently.
But Uri should be very similiar to what Pointer does. Pointer parses a C string into fragments and store them.

test/unittest/pointertest.cpp
smhdfdl: @miloyip I have removed it. It's a standard IBM thing, but we are fine if you are not comfortable with it.

include/rapidjson/uri.h
smhdfdl: @miloyip Removed.

include/rapidjson/schema.h
smhdfdl: I have made use of std::string optional. The new `GenericUri` class is now similar to `GenericPointer` class; it uses a single malloc to get all the memory it needs; there is a copy constructor and operator=; the `Resolve` methods returns a new `GenericUri` instead of modifying `this`. 

Other tings not note:
- I hit a bug (again) - issue #1805
- Added `StrCmp()` to `strfunc.h`

smhdfdl:
@miloyip I have incorporated the equivalent issue #1899 fix to the `GenericUri` class. Please can you re-review?

smhdfdl:
@miloyip Please can you re-review? I would like this merged asap so I can proceed with other PRs.

smhdfdl:
@miloyip Please can you re-review?

PR #1858, opened 2021-03-13
Provide StdAllocator, STL compatible, for use with standard types

ylavic:
This StdAllocator is a std::allocator, thus usable with the STL containers and/or std::basic_string.

It's built on top (composition) of any allocator implementing the rapidjson Allocator concept :
```
template <typename T, typename BaseAllocator = CrtAllocator>
class StdAllocator :
    public std::allocator<T>
{
   ...
    /* implicit */
    StdAllocator(const BaseAllocator& allocator) RAPIDJSON_NOEXCEPT :
        allocator_type(),
        baseAllocator_(allocator)
    { }
   ...
};
```
So wherever a rapidjson Allocator is available (i.e. almost anywhere), one can define a STL containter or string like:
```
MemoryPoolAllocator<> a;
typedef StdAllocator<Ch, MemoryPoolAllocator<> > ChAllocator;
typedef std::basic_string<Ch, std::char_traits<Ch>, ChAllocator> String;
String s{ChAllocator(a)};
s = ...;
```

This requires that the BaseAllocator be copyable, so this PR make it so for MemoryPoolAllocator by sharing/refcounting the allocated chunks between copies. I don't expect non-copyable => copyable to be a compatibility issue..
CrtAllocator is already copyable (an empty class actually), nothing to change there.

#1485 can be rebased on top of this PR (will do), and possibly #1848's GenericUri could use it too to forward the user allocator to the `String` members.

PR #1865, opened 2021-03-27
Fix StdAllocator deprecation errors in C++17 mode

sylveon:
This fixes a bug introduced in #1858 which prevented from using RapidJSON under MSVC when `/std:c++17` or `/std:c++latest` was used, because the added code used members which where deprecated in C++17 and removed in C++20. This triggered an error.


sylveon:
Replaced by #1866

PR #1866, opened 2021-03-28
Make StdAllocator C++17-20 compatible.

ylavic:
Alternatively to #1865, this proposal (also) fixes StdAllocator with regard to C++17 and later, though the choice here is to keep the same interface for StdAllocator (the C++11 and earlier one) even if compiled with newer C++ versions. Since std::allocator_traits is already available in C++11, there is no need for C++17 or C++20 specifics to allow for that.

While at it, this PR also fixes RAPIDJSON_DELIBERATE_FALLTHROUGH which is not valid with a compiler which __has_cpp_attribute(fallthrough) but still [[fallthrough]] is not usable because -std=C++98 or -std=C++11 is specified.

Sorry for the alternate proposal @sylveon, I thought it would be faster/easier to propose and show some code rather than discuss my proposed changes in your PR..

### DIFF
--- a/include/rapidjson/allocators.h
+++ b/include/rapidjson/allocators.h
@@ -19,6 +19,11 @@
 
 #include <memory>
 
+#if RAPIDJSON_HAS_CXX11
+#include <limits>
+#include <type_traits>
+#endif
+
 RAPIDJSON_NAMESPACE_BEGIN
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -420,6 +425,11 @@ class StdAllocator :
     public std::allocator<T>
 {
     typedef std::allocator<T> allocator_type;
+#if RAPIDJSON_HAS_CXX11
+    typedef std::allocator_traits<allocator_type> traits_type;
+#else
+    typedef allocator_type traits_type;
+#endif
 
 public:
     typedef BaseAllocator BaseAllocatorType;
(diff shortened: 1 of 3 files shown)


sylveon:
Sounds good to me, however I would heavily suggest to use the `RAPIDJSON_CPLUSPLUS` macro I added.

MSVC has the value of `__cplusplus` always set to `199711L` by default, so without the macro I added, this change does not resolve the issue on MSVC, since it will still hit the pre-C++ 11 codepath even when C++ 17 is enabled.

It also completely disables the `[[fallthrough]]` attribute on MSVC, for the same reason.

include/rapidjson/allocators.h
```diff
@@ -449,29 +459,50 @@ class StdAllocator :
...
+    }
+    const_pointer address(const_reference r) const RAPIDJSON_NOEXCEPT
+    {
+        return std::addressof(r);
+    }
+
     size_t max_size() const RAPIDJSON_NOEXCEPT
     {
-        return allocator_type::max_size();
+        return std::numeric_limits<size_type>::max() / sizeof(value_type);
```
sylveon, line 489: Why not do `return traits_type::max_size(*this);`? Same for `construct` and `destroy`.

ylavic: The documentation [1] seems to suggest that std::allocator_traits::max_size() calls the passed in allocator max_size() member if available, that'd be an infinite recursion I suppose here..

[1] https://en.cppreference.com/w/cpp/memory/allocator_traits/max_size

sylveon: We'd be using the allocator_traits for std::allocator<T>, not StdAllocator<T>, so it should work fine.

ylavic:
> Sounds good to me, however I would heavily suggest to use the `RAPIDJSON_CPLUSPLUS` macro I added.

OK, I'll integrate that. I saw it in your PR but don't know much about MSVC so did not copy that bits blindly. Thanks for explaining.

include/rapidjson/allocators.h
ylavic: Hm ok, `traits_type::max_size(*this)` would be using `(*this).std::allocator::max_size()`.
That's tricky though, and since std::allocator::max_size() is quite unlikely to return something else than the above I'd rather avoid it.

Given that my code does `typedef allocator_type traits_type` for C++ < 11, we could
```
    size_t max_size() const RAPIDJSON_NOEXCEPT
    {
        return traits_type::max_size();
    }
```
regardless of C++ version, but is it worth it? Some special MSVC's std::allocator::max_size() again?

sylveon: That wouldn't work because allocator_traits wants us to pass in an instance of the allocator while calling the base class's member doesn't.

I think the existing code is fine, just a nit.

ylavic: Ah yes, right. Thanks.

include/rapidjson/rapidjson.h
```diff
@@ -124,6 +124,17 @@
 #define RAPIDJSON_NAMESPACE_END }
 #endif
 
+///////////////////////////////////////////////////////////////////////////////
+// __cplusplus macro
+
+//!@cond RAPIDJSON_HIDDEN_FROM_DOXYGEN
```
sylveon, line 130: Missing an `endcond`

ylavic: Thanks, added in latest version.

include/rapidjson/allocators.h
ylavic: Finally I used the traits_type version of `max_size`, `construct` and `destroy` because MSVC somehow didn't like my explicit `p->~T()` in `destroy` (see [1]). So let it use its own implementation..

[1] https://ci.appveyor.com/project/miloyip/rapidjson-0fdqj/builds/38443196/job/m66fsm9nbyte8vvx#L247

sylveon (approved):

miloyip:
I am unfamiliar with C++17 or above. If @sylveon thinks this PR is OK, please leave a comment and I will merge it. Thank you.
Ideally we shall add appveyor config in future for testing such compilation configuration.

sylveon:
The PR is ok

ylavic:
> Ideally we shall add appveyor config in future for testing such compilation configuration.

Tried that in #1868, let's see how it goes (normally it shouldn't pass with current master until this PR is merged, then I'd pull master there and it'd pass).

PR #1868, opened 2021-03-29
Handle C++17 (and C++11 with MSVC) in CI.

ylavic:
Tentative, my cmake/travis/appveyor foo being quite limited..


ylavic:
Failing as expected with MSVC's /std:C++17.
Should pass once #1866 is merged.

appveyor.yml
```diff
@@ -13,37 +13,67 @@ environment:
...
     VS_VERSION: 15 2017
     VS_PLATFORM: x64
+    CXX11: OFF
+    CXX17: OFF
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    VS_VERSION: 15 2017
+    VS_PLATFORM: x64
+    CXX11: ON
+    CXX17: OFF
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
```
sylveon, line 68: I'd try using the VS 2019 image for C++17, it has better support

ylavic: Let's see how a8bd931 goes

PR #1904, opened 2021-06-11
Fix Pointer::Append() crash for custom allocator on Windows

### LINKED ISSUE #1899: Pointer case throws SEH exception on Windows
@miloyip  I have hit this problem again as part of the work to rewrite `Uri` so it does not use `std::string`. The rewritten `Uri` class has a method `GenericUri Resolve(const GenericUri& baseuri, Allocator* allocator = 0) {...}`. This is very similar to `Pointer` class method `GenericPointer Append(const Token& token, Allocator* allocator = 0) {...}`.  Below are minimal tests that show the problem.

This test fails with `error: SEH exception with code 0xc0000005 thrown in the test body.` only on Windows.
```
    typedef GenericUri<Value, MemoryPoolAllocator<> > UriType;
    CrtAllocator allocator;

    UriType base = UriType("http://auth/path/#frag");
    UriType ref = UriType("http://newauth/newpath#newfrag");
    UriType res = ref.Resolve(base);
    EXPECT_TRUE(StrCmp(res.GetString(), "http://newauth/newpath#newfrag") == 0);
```
This test passes on Windows.
```
    typedef GenericUri<Value> UriType;
    
    UriType base = UriType("http://auth/path/#frag");
    UriType ref = UriType("http://newauth/newpath#newfrag");
    UriType res = ref.Resolve(base);
    EXPECT_TRUE(StrCmp(res.GetString(), "http://newauth/newpath#newfrag") == 0);
```
I can make the equivalent change to one of the `PointerTest.cpp` tests for `Append`, and I get the same failure. This fails.
```
        typedef GenericPointer<Value, MemoryPoolAllocator<> > PointerType;
        PointerType p;
        PointerType q = p.Append("foo");
        EXPECT_TRUE(PointerType("/foo") == q);
        q = q.Append(1234);
        EXPECT_TRUE(PointerType("/foo/1234") == q);
        q = q.Append("");
        EXPECT_TRUE(PointerType("/foo/1234/") == q);
```
This passes (`Pointer` is defined in `pointer.cpp` as `typedef GenericPointer<Value> Pointer;`).
```
        Pointer p;
        Pointer q = p.Append("foo");
        EXPECT_TRUE(Pointer("/foo") == q);
        q = q.Append(1234);
        EXPECT_TRUE(Pointer("/foo/1234") == q);
        q = q.Append("");
        EXPECT_TRUE(Pointer("/foo/1234/") == q);
```

_Originally posted by @smhdfdl in https://github.com/Tencent/rapidjson/issues/1805#issuecomment-855247235_

miloyip:
On some compiler, when copy elision is not activated, triggers the bug.

When `Append()` is called it returns a new `Pointer`. And within internal calls of `Append()`, the `return` may call copy constructor, which incorrectly using the source's allocator. This cause problem because the source is temporary and will be destructed.

@smhdfdl

Fix #1899

### DIFF
--- a/include/rapidjson/pointer.h
+++ b/include/rapidjson/pointer.h
@@ -163,7 +163,7 @@ public:
     GenericPointer(const Token* tokens, size_t tokenCount) : allocator_(), ownAllocator_(), nameBuffer_(), tokens_(const_cast<Token*>(tokens)), tokenCount_(tokenCount), parseErrorOffset_(), parseErrorCode_(kPointerParseErrorNone) {}
 
     //! Copy constructor.
-    GenericPointer(const GenericPointer& rhs) : allocator_(rhs.allocator_), ownAllocator_(), nameBuffer_(), tokens_(), tokenCount_(), parseErrorOffset_(), parseErrorCode_(kPointerParseErrorNone) {
+    GenericPointer(const GenericPointer& rhs) : allocator_(), ownAllocator_(), nameBuffer_(), tokens_(), tokenCount_(), parseErrorOffset_(), parseErrorCode_(kPointerParseErrorNone) {
         *this = rhs;
     }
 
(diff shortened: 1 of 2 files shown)


smhdfdl:
@miloyip I have tested this locally in my product's build and the fix looks good.

xkszltl:
Sounds like a related regression, could you take a look?
- https://github.com/Tencent/rapidjson/issues/2022

PR #1930, opened 2021-08-26
Fix schema regex preprocessor include logic

bryant-ferguson:
The logic for the regex engine selection was broken and [this PR](https://github.com/Tencent/rapidjson/pull/1161) didn't really help and introduced an additional bug.

That PR gave the illusion of working, but made it impossible to disable the std::regex engine explicitly. (If you defined RAPIDJSON_SCHEMA_USE_STDREGEX=0, it would be reset to 1).

Also, previously if you were to set RAPIDJSON_SCHEMA_USE_INTERNALREGEX = 1, it would get reset to 0, disabling it. 

I've reworked the logic so that the behavior matches what is specified in the documentation. Internal is the default and overrides std::regex and if you disable internal (or both) you get no regex implementation.

Let me know what you think. It could be nice make it such that if you only explicitly set RAPIDJSON_SCHEMA_USE_STDREGEX =1, that you will get std::regex (and not internal), but my goal was bare minimum compliance with documentation.


bferg88:
@miloyip Please have a look at this.

bferguson-impinj:
@miloyip ?

djbn65:
@miloyip, @tencent-adm, any chance we can get this pushed in? It's causing unintuitive behavior in schema.h for example if both RAPIDJSON_SCHEMA_USE_INTERNALREGEX and RAPIDJSON_SCHEMA_USE_STDREGEX are defined to be 0. In that case, we end up with RAPIDJSON_SCHEMA_USE_STDREGEX getting set to 1 even though we defined it to 0 previously. Also, if RAPIDJSON_SCHEMA_USE_INTERNALREGEX is defined to be 1, it gets reset to 0 due to the !defined(RAPIDJSON_SCHEMA_USE_INTERNALREGEX) check.

PR #1944, opened 2021-10-02
Support CMake none targets

ilelann:
When trying to import rapidjson with for exemple : 
fetchcontent_declare(rapidjson GIT_REPOSITORY https://github.com/Tencent/rapidjson.git)

if your CMake/Clang is "bare metal", exemple given : 

set(CMAKE_SYSTEM_NAME none)
set(CMAKE_SYSTEM_PROCESSOR x86_64)
set(CMAKE_C_COMPILER_TARGET x86_64-elf-none)
set(CMAKE_CXX_COMPILER_TARGET x86_64-elf-none)

CMake fails to process CMakeLists.txt because of the switch on UNIX/CYGWIN/WIN32 for install directory.

Error is: 

CMake Error at cmake-build-debug-clang/_deps/rapidjson-src/CMakeLists.txt:244 (INSTALL):
  INSTALL FILES given no DESTINATION!


Crayon2000:
I was having the same error when compiling my code for PowerPC EABI. While waiting for this PR to be merged, here is a workaround. Simply add this parameter when calling camke: `-D_CMAKE_INSTALL_DIR=/.`

PR #1947, opened 2021-10-12
Making GenericMemberIterator comply with C++20's std::random_access_iterator concept

the-moisrex:
In C++20, having this test will fail:
```c++
static_assert(std::random_access_iterator<typename rapidjson_object_type::MemberIterator>);
```
Thus GCC explains it with this:
```
/usr/include/c++/11.1.0/bits/iterator_concepts.h:660:13:   required by the constraints of ‘template<class _Iter> concept std::random_access_iterator’
/usr/include/c++/11.1.0/bits/iterator_concepts.h:664:10:   in requirements with ‘_Iter __i’, ‘const _Iter __j’, ‘std::iter_difference_t<_Iter> __n’ [with _Iter = rapidjson::GenericMemberIterator<false, rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >]
/usr/include/c++/11.1.0/bits/iterator_concepts.h:669:15: note: the required expression ‘(__n + __j)’ is invalid
  669 |         { __n +  __j } -> same_as<_Iter>;
      |           ~~~~^~~~~~
```

By adding operator+, will fix the problem.


the-moisrex:
Okay I squashed the commits into one; I'm not sure about the coding styles, but I did change it to look more horrible than the original one 😄 
It would be nice to have a `.clang-format`!

PR #1950, opened 2021-10-12
Jsonstruct

abwilson:
I wrote this a few years ago. It allows rapidjson to interface with strongly typed struct-like objects. See the unit test to get an idea of what you can do with it. There are also some changes in this PR that were required to get this building on my Mac - (Catalina). I should also say that some of the code requires c++14 as a minimum.

### DIFF
--- a/test/unittest/readertest.cpp
+++ b/test/unittest/readertest.cpp
@@ -1399,7 +1399,7 @@ public:
   WIStreamWrapper(std::wistream& is) : is_(is) {}
 
   Ch Peek() const {
-    auto c = is_.peek();
+    wchar_t c = is_.peek();
     return c == std::char_traits<wchar_t>::eof() ? Ch('\0') : static_cast<Ch>(c);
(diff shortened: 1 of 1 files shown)


dinomight:
I think since this library is targeting pre-C++11 support on fairly old compilers, the C++14 requirement probably prohibits this from being merged. I'm also concerned about the lack of documentation. The tests aren't making it immediately clear what a good use case for this is and how to best utilize it. As the author, you're in the best position to make the argument for adoption of this feature and write usage documentation, so including a markdown document describing the feature may help increase interest in enabling JSON struct for RapidJSON.

miloyip:
I think this is kinda of an add-on to existing APIs.
Will you consider putting it in `/contrib`?

abwilson:
> I think this is kinda of an add-on to existing APIs.
> Will you consider putting it in `/contrib`?

I'm happy to move it to contrib.

PR #1987, opened 2022-02-08
Fix the alignment of placement new buffer for GenericValue.

adamcalhoon:
When using operator[] on a GenericValue type clang-tidy complains,
appropriately, about the alignment of the buffer used for placement-new
of the "dummy" GenericValue.

PR #1989, opened 2022-02-09
Fix the alignment of placement new buffer for GenericValue.

adamcalhoon:
When using operator[] on a GenericValue type clang-tidy complains,
appropriately, about the alignment of the buffer used for placement-new
of the "dummy" GenericValue.

See: https://github.com/Tencent/rapidjson/pull/1987

### DIFF
--- a/include/rapidjson/document.h
+++ b/include/rapidjson/document.h
@@ -1235,8 +1235,8 @@ public:
             // return NullValue;
 
             // Use static buffer and placement-new to prevent destruction
-            static char buffer[sizeof(GenericValue)];
-            return *new (buffer) GenericValue();
+            static GenericValue buffer;
+            return *new (reinterpret_cast<char *>(&buffer)) GenericValue();
         }
     }
     template <typename SourceAllocator>


adamcalhoon:
Ping. I submitted #1987 with a typo, it got merged and correctly reverted. This PR corrects the typo and matching my local changes. Apologies again for the bad initial PR.

include/rapidjson/document.h
```diff
@@ -1235,8 +1235,8 @@ class GenericValue {
             // return NullValue;
 
             // Use static buffer and placement-new to prevent destruction
-            static char buffer[sizeof(GenericValue)];
-            return *new (buffer) GenericValue();
+            static GenericValue buffer;
+            return *new (reinterpret_cast<char *>(&buffer)) GenericValue();
```
jaepil, line 1239: I don't think using placement-new here is thread-safe, and it will (rarely) cause undefined behavior. The default constructor of `GenericValue` can be executed by different threads simultaneously and each of them initializes its members on the same memory because `buffer` is declared as `static`.

PR #2023, opened 2022-04-18
fixes for natvis

jxy-s:
This adjusts two things in the `rapidjson.natvis` file. 

First, string types would display content after the string in the case where the string does not have a null terminator. In this situation the debugger may display garbage after the string. The fix for this is informing the debugger of the length of the string to display (`[data_.s.length]`).

Second, is a fix displaying for rapidjson values in windbg. The natvis must delineate between the current type scope and the type enumeration. Without this change windbg is unable to format the type property - the change here is to prefix the type enumeration with `rapidjson::` in the natvis.

Visual Studio:
image

windbg:
image


jxy-s:
Just pinging on this one. Looks like my other PR was accepted here: https://github.com/Tencent/rapidjson/pull/2034. I did sign the CLA with that one, unsure why this one didn't prompt me.

PR #2091, opened 2022-11-04
do not define operator!= in C++20

mumbleskates:
A change to the semantics of equality operator rewriting in C++20 (P2468R2: The Equality Operator You Are Looking For) means that operator== may not be rewritten with reversed operands if operator!= is also defined. Since operator!= can normally be synthesized from operator== regardless in this language standard, we can and should avoid defining those when the new language semantics are available.

This fixes the compilation of tests (and probably consuming code) in C++20 onwards for compilers that implement this new semantic, including recent nightly builds of clang-16.

Reference: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2468r2.html


mumbleskates:
bump?

mumbleskates:
thanks!

Issue #2019, opened 2022-04-12
SIMD.SkipWhitespace_SSE2 AddressSanitizer: stack-buffer-overflow


hcoona:
Reproduce steps:

1. `git clone https://github.com/Tencent/rapidjson.git && cd rapidjson`
2. `git checkout fd3dc29a && git submodule update --init --force`
3. `mkdir build && cd build`
4. `cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug -DRAPIDJSON_BUILD_THIRDPARTY_GTEST=ON -DRAPIDJSON_BUILD_CXX17=ON -DRAPIDJSON_BUILD_ASAN=ON -DRAPIDJSON_BUILD_UBSAN=ON -DRAPIDJSON_ENABLE_INSTRUMENTATION_OPT=OFF -DRAPIDJSON_HAS_STDSTRING=ON -DRAPIDJSON_USE_MEMBERSMAP=ON -DRAPIDJSON_BUILD_DOC=OFF -DRAPIDJSON_BUILD_EXAMPLES=OFF -DCMAKE_INSTALL_PREFIX=$PWD/output`
5. `ninja unittest`
6. `bin/unittest --gtest_filter=SIMD.SkipWhitespace_SSE2`

Error reporting:

```text
RapidJSON v1.1.0
Note: Google Test filter = SIMD.SkipWhitespace_SSE2
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from SIMD
[ RUN      ] SIMD.SkipWhitespace_SSE2
=================================================================
==10384==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffef7770480 at pc 0x56317a77f469 bp 0x7ffef776fce0 sp 0x7ffef776fcd0
READ of size 16 at 0x7ffef7770480 thread T0
    #0 0x56317a77f468 in _mm_load_si128(long long __vector(2) const*) /usr/lib/gcc/x86_64-linux-gnu/9/include/emmintrin.h:697
    #1 0x56317a77f468 in rapidjson_simd::SkipWhitespace_SIMD(char const*) ../include/rapidjson/reader.h:360
[...]

PR #2101, opened 2022-12-01
Fix the address sanitizer problems - part 1

smhdfdl:
This PR is intended to address the SIMD unit test sanitizer problem identified in issue #2019, and any other sanitizer problems that are being thrown up when running other RapidJSON unit tests under `valgrind` .  #2019 contains the list but in summary:

1) SIMD.SkipWhitespace  
- Invalid read of size 16 (the original subject of issue #2019)
- Conditional jump or move depends on uninitialised value(s)
- Use of uninitialised value of size 8

2) Uri.Parse_UTF16 with std::string
- Invalid read of size 32

I am seeing all these on `master` branch (Fedora 36).

### DIFF
--- a/test/unittest/simdtest.cpp
+++ b/test/unittest/simdtest.cpp
@@ -49,10 +49,12 @@ using namespace rapidjson_simd;
 #define SIMD_SUFFIX(name) name
 #endif
 
+#define SIMD_SIZE_ALIGN(n) ((size_t(n) + 15) & ~size_t(15))
+
 template <typename StreamType>
 void TestSkipWhitespace() {
     for (size_t step = 1; step < 32; step++) {
-        char buffer[1025];
+        char buffer[SIMD_SIZE_ALIGN(1025)];
         for (size_t i = 0; i < 1024; i++)
             buffer[i] = " \t\r\n"[i % 4];
         for (size_t i = 0; i < 1024; i += step)
@@ -79,7 +81,7 @@ TEST(SIMD, SIMD_SUFFIX(SkipWhitespace)) {
 
 TEST(SIMD, SIMD_SUFFIX(SkipWhitespace_EncodedMemoryStream)) {
     for (size_t step = 1; step < 32; step++) {
-        char buffer[1024];
+        char buffer[SIMD_SIZE_ALIGN(1024)];
         for (size_t i = 0; i < 1024; i++)
             buffer[i] = " \t\r\n"[i % 4];
         for (size_t i = 0; i < 1024; i += step)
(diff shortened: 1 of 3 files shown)


smhdfdl:
Commit (a7ce844)[https://github.com/Tencent/rapidjson/commit/a7ce8449431671bc01241c7ed0cf51b6bbe4deb3] addresses the main SIMD test sanitizer problem. It also changes Uri test to pass in an allocator where it was missing (it does not fix the Uri sanitizer problem).

smhdfdl:
@ylavic  @miloyip  @hcoona are you able to help track down the remaining problems please?

smhdfdl:
Update on the valgrind failures in Uri.Parse_UTF16. The cause is nothing to do with the Uri class. I can recreate the problem using the following test case. The test passes, but valgrind thinks there is a problem with the compare of the underlying memory. It only happens when the string being compared is less than 8 chars.
```
TEST(Uri, Parse_UTF16_Std) {
    typedef GenericValue<UTF16<> > Value16;
    typedef std::basic_string<Value16::Ch> String;

    String short1 = L"abcd";
    String short2 = L"abcd";
    EXPECT_TRUE(short1 == short2);
}
```

miloyip:
> Update on the valgrind failures in Uri.Parse_UTF16. The cause is nothing to do with the Uri class. I can recreate the problem using the following test case. The test passes, but valgrind thinks there is a problem with the compare of the underlying memory. It only happens when the string being compared is less than 8 chars.
> 
> ```
> TEST(Uri, Parse_UTF16_Std) {
>     typedef GenericValue<UTF16<> > Value16;
>     typedef std::basic_string<Value16::Ch> String;
> 
>     String short1 = L"abcd";
>     String short2 = L"abcd";
>     EXPECT_TRUE(short1 == short2);
> }
> ```

It looks like a false alarm. This piece of code seems just using C++ standard library.

smhdfdl:
Commit [417bf54](https://github.com/Tencent/rapidjson/pull/2101/commits/417bf54a0ee6d18fa25de0dcf34451ebd1d6d1a8) suppresses the failure from uritest.cpp. Suspect there's an optimisation going on that is fooling valgrind into thinking there is a real problem.

smhdfdl:
That leaves SIMD.SkipWhitespace other failures ('Conditional jump or move depends on uninitialised value(s)' & 
'Use of uninitialised value of size 8'). I can't see what is wrong with the code by inspection.

smhdfdl:
@miloyip Your thoughts please

smhdfdl:
@miloyip Your thoughts please

smhdfdl:
@miloyip I agree, let's review & merge this first.

PR #2140, opened 2023-04-11
Check for __GNUC__ definition

AlbertHungGarmin:
Wrap code checking against __GNUC__ to ensure it is defined. This can cause errors with compilers which do not define this preprocessor value.


sembooooo (approved):

AlbertHungGarmin:
@miloyip, would you take a look at this patch that is beneficial to non-GNUC users?

PR #2144, opened 2023-04-20
Add curly braces around statement

AlbertHungGarmin:


sembooooo:
Can I  support in this review ?

sembooooo (approved):
The author added curly braces properly. Looks good.

sembooooo:
So does this mean i supported rapidJSON in any way @miloyip  ? (even if it is in the size of a water drop in a ocean)

AlbertHungGarmin:
@miloyip , Would you take a look at this patch? The GCC ARM compiler is generating warning for 'if' statement without curly braces.

PR #2196, opened 2023-08-24
Support for optional parsing of hexadecimals

rokj-isystem:


pagict:
please don't commit your project configurations(`.vs`) and build directories(`.out`).

PR #2199, opened 2023-09-01
Fix MSVC build

bobbymcr:
- Remove unsupported /std:c++11 flag (only c++14 and above is accepted)
- Enable MSVC runtime library flag based on CMP0091
- Ignore .vs folder
- Add extra assertions in file open tests to detect bad pointers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ Doxyfile
 Doxyfile.zh-cn
 DartConfiguration.tcl
 *.nupkg
+/.vs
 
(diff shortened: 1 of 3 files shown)


miloyip:
Seems this will break older VC versions?

bobbymcr:
> Seems this will break older VC versions?

Which version did you see it break? As far as I know this /std arg is only available starting in VS 2017 and only vaiues "c++14" and above options are supported. However I don't know for sure.

PR #2207, opened 2023-09-28
Fix object hashing in schema

### LINKED ISSUE #1558: Hasher's offset basis high and low values swapped
FNV-1a algorithm specifies the 64-bit offset basis as 0xcbf29ce484222325
https://en.wikipedia.org/wiki/Fowler–Noll–Vo_hash_function

In the code it's written as RAPIDJSON_UINT64_C2(0x84222325, 0xcbf29ce4). 
https://github.com/Tencent/rapidjson/blob/2648a732dbb8c7fa211e41616071202bfbd09c77/include/rapidjson/schema.h#L294

Shouldn't these high and low values be swapped?

### LINKED ISSUE #2205: JSON schema incorrect validation with array with uniqueItems
Tested against the release 1.1.0 and the latest master at the time of writing.

With the schema:

```json
{
  "$schema": "http://json-schema.org/schema#",
  "id": "http://test.com/schemas/test.schema.json",
  "type": "object",
  "properties": {
    "myArray": {
      "type": "array",
      "uniqueItems": true,
      "items": {
        "type": "object"
      }
    }
  }
}
```

And the JSON document:

```json
{
  "myArray": [
    {
      "a": "a"
    },
    {
      "b": "b"
    },
    {
      "c": "c"
    }
  ] 
} 
```

I get the following error from the schemavalidator example:

```text
Input JSON is invalid.
Invalid schema: #/properties/myArray
Invalid keyword: uniqueItems
Invalid code: 11
Invalid message: Array has duplicate items at indices '%duplicates' but 'uniqueItems' is true.
Invalid document: #/myArray/1
Error report:
{
    "uniqueItems": {
        "duplicates": [
            0,
            1
        ],
        "errorCode": 11,
        "instanceRef": "#/myArray",
        "schemaRef": "#/properties/myArray"
    }
}
Error Name: uniqueItems
Message: Array has duplicate items at indices '0,1' but 'uniqueItems' is true.
Instance: #/myArray
Schema: #/properties/myArray
```

But these objects are unique. If I change the first object in the array that it's saying is a duplicate:

```json
{
  "myArray": [
    {
      "a": "a1"
    },
    {
      "b": "b"
    },
    {
      "c": "c"
    }
  ]
}
```

I get an error with the next 2 elements instead:

```text
Input JSON is invalid.
Invalid schema: #/properties/myArray
Invalid keyword: uniqueItems
Invalid code: 11
Invalid message: Array has duplicate items at indices '%duplicates' but 'uniqueItems' is true.
Invalid document: #/myArray/2
Error report:
{
    "uniqueItems": {
        "duplicates": [
            1,
            2
        ],
        "errorCode": 11,
        "instanceRef": "#/myArray",
        "schemaRef": "#/properties/myArray"
    }
}
Error Name: uniqueItems
Message: Array has duplicate items at indices '1,2' but 'uniqueItems' is true.
Instance: #/myArray
Schema: #/properties/myArray
```

This same situation occurs with longer keys/values as well, which is where I came across the issue.

aikawayataro:
The objects with key-value pairs where key=value will always produce zero-value hash.
The algorithm used for keys and string values is the same, so `Hasher::Hash` called on them in `Hasher::EndObject` will produce zero-value, due to xoring in FNV-1a algorithm. Such hash is bug-prone and easy to duplicate, for example: objects
```json
{"value": "value"}
```
and
```json
{"value1": "value1", "value2": "value2" }
```
will produce same hash: `0`

Fixes #2205, fixes #1558

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -407,7 +407,7 @@ private:
     
     bool WriteBuffer(Type type, const void* data, size_t len) {
         // FNV-1a from http://isthe.com/chongo/tech/comp/fnv/
-        uint64_t h = Hash(RAPIDJSON_UINT64_C2(0x84222325, 0xcbf29ce4), type);
+        uint64_t h = Hash(RAPIDJSON_UINT64_C2(0xcbf29ce4, 0x84222325), type);
         const unsigned char* d = static_cast<const unsigned char*>(data);
         for (size_t i = 0; i < len; i++)
             h = Hash(h, d[i]);
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -367,7 +367,9 @@ public:
         uint64_t h = Hash(0, kObjectType);
         uint64_t* kv = stack_.template Pop<uint64_t>(memberCount * 2);
         for (SizeType i = 0; i < memberCount; i++)
-            h ^= Hash(kv[i * 2], kv[i * 2 + 1]);  // Use xor to achieve member order insensitive
+            // Issue #2205
+            // Hasing the key to avoid key=value cases with bug-prone zero-value hash
+            h ^= Hash(Hash(0, kv[i * 2]), kv[i * 2 + 1]);  // Use xor to achieve member order insensitive
         *stack_.template Push<uint64_t>() = h;
         return true;
     }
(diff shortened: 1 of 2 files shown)


miloyip:
Suggest to add an unit test to reproduce the issue and verify the fix.

PR #2211, opened 2023-10-18
Optinally allow parsing of hexadecimals

rokj-isystem:
Allow parsing of hexadecimals into integers, when flag `kParseHexadecimalsFlag` is set.


include/rapidjson/reader.h
```diff
@@ -1483,14 +1484,85 @@ class GenericReader {
         // Parse minus
         bool minus = Consume(s, '-');
 
-        // Parse int: zero / ( digit1-9 *DIGIT )
+        // Parse int: ( digit1-9 *DIGIT ) / hexadecimal
         unsigned i = 0;
         uint64_t i64 = 0;
         bool use64bit = false;
         int significandDigit = 0;
+        bool cont = true;
```
rokj-isystem, line 1492: Moved, because when parsing hexadecimals this declaration would be skipped otherwise.

PR #2225, opened 2023-12-05
Change an array to a vector

EstherWx:
Member 'IterativeParsingReaderHandler::Logs' is not initialized in the constructor and it's hard to initialize an array in the old version of Vsiual Studio. Change it to a vector and remove the need for LogCapacity and LogCount as separate members.

Move the const static variables out from the template struct 'IterativeParsingReaderHandler' to solve 'undefined reference for LOG_*' link errors.

### DIFF
--- a/test/unittest/readertest.cpp
+++ b/test/unittest/readertest.cpp
@@ -20,6 +20,7 @@
 #include "rapidjson/memorystream.h"
 
 #include <limits>
+#include <vector>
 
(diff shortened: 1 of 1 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/rapidjson?pullRequest=2225) <br/>All committers have signed the CLA.

PR #2235, opened 2023-12-12
Fix comparision of two doubles

EstherWx:
One of multipleOf test failed because most floating-point numbers end up being slightly imprecise. And, the check of multipleOf with two double numbers using the dividend (a) and the result of multiplying the multiple after rounding down by the divisor (floor(|a|/|b|)*b) to compare.

Change to using std::numeric_limits::epsilon to check the error of the division result.


djbn65:
@miloyip @EstherWx, any chance we could get this merged in? On our system, this fixes the .0075 multipleOf .0001 unit test for schema.h (I'm assuming different systems round doubles differently).

djbn65:
@tencent-adm, can we get this reviewed as soon as possible?

PR #2244, opened 2023-12-25
fix: delete validator before clear

junwha0511:
Hi, we found an incorrect use of Clear API of the allocator, which leads to Use-After-Free in schematest.

At test/perftest/schematest.cpp:208, the validator is constructed with validatorAllocator, and the chunks allocated during line 210 to line 213 are collected by the Clear API at line 214 as the for statement is ended. 
However, the destructor of GeneriSchemaValidator is called at line 215, leading to invalid accesses to the chunks(Stack). 

To avoid this issue, I added explicit delete to the validator before the Clear API call.

Thank you very much for taking the time to revise this issue!


miloyip:
I think may just add a block `{}` to force destructing the validator first.

junwha0511:
Hi @miloyip, Thank you for the feedback! I changed the patch as you commented.

PR #2293, opened 2024-07-23
Fix warning C4127S

SilverPlate3:
When building on Windows with C++ 17 and above we get the warning C4127S
image

This PR fixes it as it uses `if constexpr` when C++ 17 and above


SilverPlate3:
@miloyip Hope you are doing well.
When can I expect this PR to be pushed? 
Thanks!

SilverPlate3:
@tencent-adm @miloyip 
Hey.
Is this going to be pushed?
I want to remove the local repo fork, and it will delete this PR.

PR #2346, opened 2025-06-27
Correct the behaviors of WriteBuffer

EstherWx:
WriteBuffer attempt to dereference a pointer without any check. Adds a safeguard to prevent dereferencing a null pointer to fix it and nullPointer warning detected by cppcheck.

Besides, WriteBuffer always does hashing and returns true. According to the reference of FNV-1a for Hash algorithm, the original code uses a while loop to iterate over the buffer, checking that the current position is less than the end. WriteBuffer should not do hashing when the length is 0.

### DIFF
--- a/include/rapidjson/schema.h
+++ b/include/rapidjson/schema.h
@@ -357,7 +357,11 @@ public:
     }
 
     bool String(const Ch* str, SizeType len, bool) {
-        WriteBuffer(kStringType, str, len * sizeof(Ch));
+        if (len == 0) {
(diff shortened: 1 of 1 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/rapidjson?pullRequest=2346) <br/>All committers have signed the CLA.

admorgan (commented):
The second commit fixes a bug introduced in the first. It would be better to combine them into one commit.

PR #2362, opened 2025-12-14
Making GenericMemberIterator comply with C++20's std::random_access_iterator concept 

ste-lam:
This PR addresses the same issue with std::random_access_iterator as #1947.
The only difference is that this approach uses a friend operator overload.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/rapidjson?pullRequest=2362) <br/>All committers have signed the CLA.

PR #2382, opened 2026-05-14
Fix ParseNumber integer parse bug around max int32 and uint32.

NonRadical:
For integers in a range above from and including max int32 as well as a range down from and including max uint32, the current implementation of ParseNumber can "miss" that the accumulated 32 bit integer will overflow due to the evaluation of the potential overflow happening before the accumulation is done. Example values that experience this issue are 2,147,483,648 (max int32) and 4,294,967,295 (max uint32).


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/rapidjson?pullRequest=2382) <br/>All committers have signed the CLA.

PR #2386, opened 2026-06-11
Respect riscv targets in platform macros

carlosqwqqwq:
## Why

RapidJSON is header-only, but some platform macros and test/performance entry points still infer target capabilities from host compiler predefined macros. In a simulated or cross-target `riscv64` configuration built from an `x86_64` host compiler, that can incorrectly preserve x86-only assumptions such as the 48-bit pointer optimization or host SIMD auto-detection. This patch makes the generic RISC-V path explicit and keeps simulated validation builds from inheriting host-specific SIMD behavior.

## What changed

- Add a conservative `RAPIDJSON_RISCV` platform macro in `rapidjson.h`.
- Treat `__riscv_xlen == 64` as a 64-bit target in `rapidjson.h`.
- Disable `RAPIDJSON_48BITPOINTER_OPTIMIZATION` on RISC-V so RISC-V builds do not inherit the x86-64-specific pointer compression assumption.
- Update `test/perftest/perftest.h` and `test/unittest/simdtest.cpp` so simulated or cross-target RISC-V validation does not auto-enable host `SSE2`/`SSE4.2`/`NEON` paths.
- Adjust the tutorial example to use a dynamically built `std::string`, avoiding an x86-specific short-string layout assumption during non-x86 validation builds.

## Verification

- Ran native CMake configuration with Ninja using `RAPIDJSON_BUILD_DOC=OFF` and `RAPIDJSON_BUILD_TESTS=OFF`.
- Built the native examples successfully with `cmake --build build-codex-native-058 --parallel 4`.
- Ran simulated `riscv64` CMake configuration with `CMAKE_SYSTEM_NAME=Linux`, `CMAKE_SYSTEM_PROCESSOR=riscv64`, `CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY`, and forced `__riscv=1` / `__riscv_xlen=64` because no real RISC-V toolchain is available on the host.
- Preprocessed `rapidjson.h` under forced `__riscv=1` / `__riscv_xlen=64` and confirmed `RAPIDJSON_RISCV=1`, `RAPIDJSON_64BIT=1`, and `RAPIDJSON_48BITPOINTER_OPTIMIZATION=0`.
- Built the simulated example sources to object files, including `example/tutorial/tutorial.cpp.o`, and confirmed `build.ninja` contains no `-msse*`, `-mavx*`, or `NEON` flags.

## Notes

This is a conservative portability patch. It does not add a RISC-V SIMD backend. Because the host environment uses a Windows-native compiler while simulating a Linux `riscv64` target, the final executable link step still inherits a host/toolchain mismatch around `-rdynamic`. Verification therefore focuses on native example build stability, simulated RISC-V configuration, object compilation, macro checks, and x86/ARM SIMD flag exclusion rather than a full real `riscv64` link.

### DIFF
--- a/example/tutorial/tutorial.cpp
+++ b/example/tutorial/tutorial.cpp
@@ -4,6 +4,7 @@
 #include "rapidjson/document.h"     // rapidjson's DOM-style API
 #include "rapidjson/prettywriter.h" // for stringify JSON
 #include <cstdio>
+#include <string>
 
 using namespace rapidjson;
 using namespace std;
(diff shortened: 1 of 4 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/rapidjson?pullRequest=2386) <br/>All committers have signed the CLA.

PR #2398, opened 2026-09-17
Preserve RapidJSON_DIR during package discovery

### LINKED ISSUE #2376: The second find_package will cause error
https://github.com/Tencent/rapidjson/blob/24b5e7a8b27f42fa16b96fc70aade9106cf7102f/RapidJSONConfig.cmake.in#L10-L12

cmake 用`<PackageName>_DIR` 存储configuration file的目录。 RapidJSON_DIR被覆盖后会报错。

Shubham-Padkonde:
Fixes #2376.

An installed RapidJSONConfig.cmake overwrites `RapidJSON_DIR` with the producer's build directory. A second `find_package(RapidJSON CONFIG REQUIRED)` then searches that directory and fails to include RapidJSON-targets.cmake. Remove the assignment so CMake retains the discovered package location.

Add a standalone CMake consumer regression under `test/cmake/repeated_find_package`. Configure it with `-DRapidJSON_DIR=<installation>/lib/cmake/RapidJSON`, build it, and run `repeated_find_package`. It checks that discovery preserves the directory, finds the same configuration twice, and still supports compiling and linking against the imported target.

Validation on Linux, GCC 11.4 and CMake 4.4.3 (with CMAKE_POLICY_VERSION_MINIMUM=3.5 for the existing project):
- Installed-package consumer: second discovery fails on original master; configure, build, and executable pass with the fix.
- Added consumer regression: configure, build, and executable pass.
- All 469 unit tests and all 60 performance tests pass.
- `git diff --check` passes.

The open broader CMake refactor #2340 also touches this variable but still replaces its value. This change is limited to preserving CMake's discovery state.

Prepared with OpenAI Codex assistance.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/rapidjson?pullRequest=2398) <br/>All committers have signed the CLA.

Shubham-Padkonde:
Checked the AppVeyor failure: the VS2019 jobs stop during configuration at example/CMakeLists.txt:1 because the installed CMake removed compatibility below 3.5. The preceding PR build 1.1.0.1055 fails at the same line with the same message. This occurs before the installed-package consumer is exercised. The other 14 jobs in build 1.1.0.1056 passed. I have kept the package-directory fix scoped to the reported issue.

